PR #17, opened 2016-11-23
client: add 'service session-pool-update' command

melor:


saaros:
Please fix the commit title (session vs connection)

PR #23, opened 2017-02-17
cli: update database listing to match api change

melor:


saaros:
These changes haven't been rolled out to production.

PR #32, opened 2017-10-02
Cert download and cleanups

Ormod:


aiven/client/cli.py
```diff
@@ -549,6 +549,29 @@ def service_user_list(self):
     @arg.project
     @arg.service_name
     @arg("--username", help="Service user username", required=True)
+    @arg("--target-directory", help="Directory to write credentials to", required=False, default=os.getcwd())
```
melor: Add a short option alias `-d`

aiven/client/cli.py
```diff
@@ -549,6 +549,29 @@ def service_user_list(self):
...
+    def service_user_creds_download(self):
+        """Download service user certificate/key/CA certificate"""
+        project_name = self.get_project()
+        try:
+            result = self.client.get_project_ca(project=project_name)
+            with open(os.path.join(self.args.target_directory, "ca.pem"), "w") as fp:
+                fp.write(result["certificate"])
+        except client.Error as ex:
+            print(ex.response.text)
```
melor: You could just skip this `print()` and add the `ex.response.text` to the exception msg

PR #34, opened 2017-11-29
cli: Fix usage of positional arguments containing dashes

mwfrojdman:
argparse turns dashes in options into underscores in `Namespace`
attribute names, but in positional arguments they stay as dashes. Fix
parsed argument references to use dashes to avoid `AttributeError`
exceptions.


mwfrojdman:
Tested locally, avn card update and delete now work

PR #45, opened 2018-03-28
avn service connect command, which runs psql with correct arguments

gugu:
I found command `gcloud compute ssh server-name` useful, so I implemented `avn service connect` command, which does exactly the same. Only for postgres for now as I don't have other services to test (and don't need them)


saaros:
https://github.com/aiven/aiven-client/pull/47 implemented this in a bit more generic fashion.

PR #46, opened 2018-04-11
client: Add command to create Java keystore/truststore/client properties

Ormod:
This makes the use of Java based clients much easier as we create
a suitable keystore/truststore/client properties file for immediate
use.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -12,6 +12,7 @@ import getpass
 import json as jsonlib
 import os
 import requests
+import subprocess
 import time
 
 PLUGINS = []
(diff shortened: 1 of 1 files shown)


aiven/client/cli.py
```diff
@@ -519,13 +520,63 @@ def service_user_list(self):
...
+        subprocess.check_call([
+            "openssl", "pkcs12", "-export",
+            "-out", os.path.join(self.args.target_directory, "client.keystore.p12"),
+            "-inkey", os.path.join(self.args.target_directory, "service.key"),
+            "-in", os.path.join(self.args.target_directory, "service.cert"),
+            "-certfile", os.path.join(self.args.target_directory, "ca.pem"),
+            "-passout", "pass:{}".format(self.args.password),
+        ])
+        service = self.client.get_service(project=self.get_project(), service=self.args.name)
+        with open(os.path.join(self.args.target_directory, "client.properties"), "w") as fp:
```
hnousiainen: This is Kafka specific, should the command name then reflect that this is intended for Kafka?

ojarva: Why is this Kafka specific? Doesn't the same apply for any other service with Java client and client certs, such as Cassandra beta?

hnousiainen: Cassandra users password authentication, so at least the credentials are different. Truststore is probably generated for server side authentication.

client.properties is surely Kafka specific.

Ormod: Updated name to be Kafka specific.

PR #51, opened 2018-05-21
service: select project vpc

mwfrojdman:
Project VPC's can be specified for service create and update.

`avn service create --cloud <cloud name> --project-vpc-id <UUID>` will
place the service into a project VPC. Note that the VPC has to be in the
same cloud as the service.

`avn service create --no-project-vpc` puts the service into a public
network in the selected cloud, avoiding the default behavior where the
service would be put into a project VPC if one exists for the project in
that cloud. If there is no VPC in the selected cloud, the argument has
no effect.

The arguments have the same meaning for `avn service update`, except
that omitting the arguments will leave the service in the network it
currently is in (unless a cloud migration is initiated using `--cloud`).

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -1198,6 +1198,16 @@ ssl.truststore.type=JKS
             print(ex.response.text)
             raise
 
+    def _get_service_project_vpc_id(self):
+        """Utility method for service_create and service_update"""
+        if self.args.project_vpc_id is None:
+            project_vpc_id = None if self.args.no_project_vpc else client.UNDEFINED
+        elif self.args.no_project_vpc:
+            raise argx.UserError("Only one of --project-vpc-id and --no-project-vpc can be specified")
+        else:
+            project_vpc_id = self.args.project_vpc_id
+        return project_vpc_id
+
     @arg.project
     @arg.service_name
     @arg("--group-name", help="service group", default="default")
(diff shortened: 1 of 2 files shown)


aiven/client/client.py
```diff
@@ -22,6 +22,9 @@ def __init__(self, response, status=520):
         self.status = status
 
 
+UNDEFINED = object()
```
Ormod: Just use None for this.

aiven/client/cli.py
```diff
@@ -1198,6 +1198,21 @@ def vpc__peering_connection__request(self):
...
+        """Utility method for service_create and service_update"""
+        if self.args.project_vpc_id is None:
+            project_vpc_id = None if self.args.no_project_vpc else client.UNDEFINED
+        elif self.args.no_project_vpc:
+            raise argx.UserError("Only one of --project-vpc-id and --no-project-vpc can be specified")
+        else:
+            project_vpc_id = self.args.project_vpc_id
+        return project_vpc_id
+
+    _service_project_vpc_id_help = "Put service into a project VPC. The VPC's cloud must match the service's cloud"
```
Ormod: I wouldn't mind you just using these helps in two places as copypaste, they're unlikely to change and would match the rest of the methods.

aiven/client/client.py
mwfrojdman: There are three options for the `project_vpc_id` property in the HTTP API:
- Setting it to a UUID (a result of using `--project-vpc-id`)
- Setting it to `null` (a result of using `--no-project-vpc`)
- Not defining the property (a result of not using either of the switches)

The `UNDEFINED` constant is used for the third case - a None value leads to the second (`null`), which has a different meaning. This way is backwards compatible.

aiven/client/cli.py
mwfrojdman: Changed into passing identical string literals and removing the class attributes

PR #52, opened 2018-05-24
cli: Add support for new service logs functionality

Ormod:
This supports both fetching of all historical logs of a service that
exists (typically 24 hours) and to tail the service logs for
new entries forever.


rikonen:
Passing `--follow` parameter should automatically set sort order to `asc` because otherwise the flag doesn't really have any effect.

PR #67, opened 2018-10-19
show default values in help for integers and non-empty strings

harmti:
For example:
```
> python -m aiven.client service logs --help
...
   -S {desc,asc}, --sort-order {desc,asc}
                        Sort direction for log fetching (default: asc)
   -n LIMIT, --limit LIMIT
                        Get up to N rows of logs (default: 100)
...
```

### DIFF
--- a/aiven/client/argx.py
+++ b/aiven/client/argx.py
@@ -33,6 +33,20 @@ class CustomJsonEncoder(jsonlib.JSONEncoder):
         return jsonlib.JSONEncoder.default(self, o)
 
 
+class CustomFormatter(argparse.RawDescriptionHelpFormatter):
+    """Help formatter to display the default value only for integers and non-empty strings"""
+    def _get_help_string(self, action):
+        help_text = action.help
+        if '%(default)' not in action.help and action.default is not argparse.SUPPRESS:
+            if action.option_strings or action.nargs in [argparse.OPTIONAL, argparse.ZERO_OR_MORE]:
+                if (
+                        (not isinstance(action.default, bool) and isinstance(action.default, int))
+                        or (isinstance(action.default, str) and action.default)
+                ):
+                    help_text += ' (default: %(default)s)'
+        return help_text
+
+
 class UserError(Exception):
     """User error"""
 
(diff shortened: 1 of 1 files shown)


aiven/client/argx.py
```diff
@@ -33,6 +33,22 @@ def default(self, o):  # pylint:disable=E0202
         return jsonlib.JSONEncoder.default(self, o)
 
 
+class CustomFormatter(argparse.ArgumentDefaultsHelpFormatter, argparse.RawDescriptionHelpFormatter):
+    """Help formatter to display the default value only for integers and non-empty strings"""
+    def _get_help_string(self, action):
```
mwfrojdman: Why is `ArgumentDefaultsHelpFormatter` inherited? It only defines the `_get_help_string` method, which this class overrides. Both superclasses inherit from `HelpFormatter`.

aiven/client/argx.py
```diff
@@ -33,6 +33,22 @@ def default(self, o):  # pylint:disable=E0202
...
+        help_text = action.help
+        if '%(default)' not in action.help:
+            if action.default is not argparse.SUPPRESS:
+                defaulting_nargs = [argparse.OPTIONAL, argparse.ZERO_OR_MORE]
```
mwfrojdman: What's `defaulting`? Could just be inline in the `if` statement below?

aiven/client/argx.py
```diff
@@ -33,6 +33,22 @@ def default(self, o):  # pylint:disable=E0202
...
+                if action.option_strings or action.nargs in defaulting_nargs:
+                    if (
```
mwfrojdman: All four nested if statements could be combined into one. If that is too unwieldy, consider using `if all([<first if condition>, <second if condition>, <third>, <fourth>]):`

aiven/client/argx.py
mwfrojdman: Never mind, misread that and understood that it's a default for `nargs`, not if a default help text should be used for certain `nargs` values.

aiven/client/argx.py
harmti: Dropped ArgumentDefaultsHelpFormatter

aiven/client/argx.py
harmti: Simplified, but still 3 separate if clauses left

PR #78, opened 2019-02-05
vpc: cross region peering connection support

mwfrojdman:


aiven/client/cli.py
```diff
@@ -1313,15 +1313,54 @@ def vpc__peering_connection__list(self):
...
             raise argx.UserError(msg)
 
-    def _vpc_peering_connection_create(self):
+    @arg.project
+    @arg("--project-vpc-id", required=True, help=_project_vpc_id_help)
+    @arg("--peer-cloud-account", required=True, help="AWS account ID or Google project ID")
+    @arg("--peer-vpc", required=True, help="AWS VPC ID or Google VPC network name")
+    @arg.json
+    @arg.verbose
+    def vpc__peering_connection__get(self):
```
ojarva: describe/show would be more in line with other commands but :man_shrugging: We have a few `get` commands as well, mostly for very simple things, such as `acl get` and `system config get`.

PR #84, opened 2019-03-06
Service plans: filter by service type and show monthly prices

saaros:


ojarva:
`aiven/client/cli.py:1500:8: R1720: Unnecessary "elif" after "raise" (no-else-raise)` - we might want to disable this check altogether.

Issue #81, opened 2019-02-22
No Kafka Connect or Schema Registry functionality


slepkin:
I'm sure this is on your roadmap, but I just wanted to note that it'd be very nice to have the Kafka Connect and Schema Registry [API endpoints](https://api.aiven.io/doc/#api-Service__Kafka-ServiceSchemaRegistryCompatibility) accessible via the CLI.

PR #94, opened 2019-04-26
Add support for Kafka connect connectors.

fmorato:
Addresses #81.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -1137,6 +1137,140 @@ ssl.truststore.type=JKS
                             table_layout=layout)
 
     @arg.project
+    @arg.service_name
+    def service_connector_available(self):
+        """List available Kafka connectors"""
+        project_name = self.get_project()
+
+        response = self.client.get_available_kafka_connectors(project_name, self.args.name)
+
+        self.print_response(response)
+
+    @arg.project
+    @arg.service_name
+    def service_connector_list(self):
+        """List Kafka connectors"""
+        project_name = self.get_project()
+
+        response = self.client.list_kafka_connectors(project_name, self.args.name)
+
+        self.print_response(response)
+
+    @arg.project
+    @arg.service_name
+    @arg("connector", help="Connector name")
+    def service_connector_status(self):
+        """Get Kafka connector status"""
+        project_name = self.get_project()
+
+        response = self.client.get_kafka_connector_status(project_name, self.args.name, self.args.connector)
+
+        self.print_response(response)
+
+    @arg.project
+    @arg.service_name
+    @arg("connector", help="Connector name")
+    def service_connector_configuration(self):
+        """Get Kafka connector configuration"""
+        project_name = self.get_project()
+
+        response = self.client.get_kafka_connector_configuration(project_name, self.args.name, self.args.connector)
+
+        self.print_response(response)
+
+    def get_connector_config(self, path_or_string):
+        config = path_or_string
+        if os.path.isfile(config):
+            with open(config, 'r') as config_file:
+                config = config_file.read()
+        return config
+
+    @arg.project
+    @arg.service_name
+    @arg("connector_config", help="Connector configuration as a json file or string")
+    def service_connector_create(self):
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -1136,6 +1136,140 @@ def service_acl_list(self):
...
+        project_name = self.get_project()
+
+        response = self.client.get_kafka_connector_configuration(project_name, self.args.name, self.args.connector)
+
+        self.print_response(response)
+
+    def get_connector_config(self, path_or_string):
+        config = path_or_string
+        if os.path.isfile(config):
+            with open(config, 'r') as config_file:
```
rikonen: Prefer `"` over `'` unless there's a specific reason not to (e.g. because the string itself contains `"` characters.

aiven/client/cli.py
```diff
@@ -1136,6 +1136,140 @@ def service_acl_list(self):
...
+                config = config_file.read()
+        return config
```
rikonen: Would be good to do `json.loads` here to ensure the config actually is valid JSON.

aiven/client/cli.py
```diff
@@ -1136,6 +1136,140 @@ def service_acl_list(self):
...
+
+        self.print_response(response)
+
+    @arg.project
+    @arg.service_name
+    @arg("connector", help="Connector name")
+    def service_connector_configuration(self):
+        """Get Kafka connector configuration"""
+        project_name = self.get_project()
+
```
rikonen: The generic style of existing code seems to be fairly dense and not having line break here or after the next line would seem to be more consistent.

aiven/client/cli.py
```diff
@@ -1136,6 +1136,140 @@ def service_acl_list(self):
...
+    @arg.project
+    @arg.service_name
+    @arg("connector", help="Connector name")
+    def service_connector_delete(self):
+        """Delete a Kafka connector"""
+        project_name = self.get_project()
+
+        response = self.client.delete_kafka_connector(project_name, self.args.name, self.args.connector)
+
+        self.print_response(response["message"])
```
rikonen: Does `self.print_response` actually work for simple strings? I thought it requires `dict` or list of `dict`s.

aiven/client/cli.py
```diff
@@ -1136,6 +1136,140 @@ def service_acl_list(self):
...
+    @arg("task", help="Task id")
+    def service_connector_restart_task(self):
+        """Restart a Kafka connector task"""
+        project_name = self.get_project()
+
+        response = self.client.restart_kafka_connector_task(
+            project_name,
+            self.args.name,
+            self.args.connector,
+            self.args.task)
```
rikonen:
```
            self.args.task)
```

=>

```
            self.args.task,
        )
```

would be according to the style used in this project. Or alternatively

```
        response = self.client.restart_kafka_connector_task(
            project_name, self.args.name, self.args.connector, self.args.task
        )
```

aiven/client/client.py
```diff
@@ -349,6 +349,43 @@ def add_service_kafka_acl(self, project, service, permission, topic, username):
...
+
+    def create_kafka_connector(self, project, service, config):
+        return self.verify(self.post, "/project/{}/service/{}/connectors".format(project, service), body=config)
+
+    def update_kafka_connector(self, project, service, connector_name, config):
+        url = "/project/{}/service/{}/connectors/{}"
+        return self.verify(self.put, url.format(project, service, connector_name), body=config)
+
+    def delete_kafka_connector(self, project, service, connector_name):
+        return self.verify(self.delete, "/project/{}/service/{}/connectors/{}".format(project, service, connector_name))
```
rikonen: URL encoding in existing code is handled badly so this is not worse than existing code as such but connector names are known to be able to contain any characters so need to have `connector_name` => `quote(connector_name, safe="")`

aiven/client/cli.py
fmorato: It does work but I will remove the printing to be more consistent with the rest of the code.

PR #95, opened 2019-04-26
Add argument to display version.

### LINKED ISSUE #87: Add --version argument
It would be useful to have a `--version` argument for `avn`. Currently, if I want to know which client version I'm using, I have to execute `pip show aiven-client`.

fmorato:


fmorato:
`version.py` is generated when the package is installed and tests are not installing it before running tests.

aiven/client/argx.py
```diff
@@ -105,6 +111,8 @@ def __init__(self, name):
         self.parser = argparse.ArgumentParser(prog=name, formatter_class=CustomFormatter)
         self.parser.add_argument("--config", help="config file location %(default)r",
                                  default=envdefault.AIVEN_CLIENT_CONFIG)
+        if __version__:
```
ojarva: Isn't this always true, either with real version or with "UNKNOWN" placeholder?

fmorato: Indeed.

PR #97, opened 2019-05-03
Add support for Kafka Schema Registry

fmorato:
Addresses #81

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -1299,6 +1299,141 @@ ssl.truststore.type=JKS
         self.client.restart_kafka_connector_task(project_name, self.args.name, self.args.connector, self.args.task)
 
     @arg.project
+    @arg.service_name
+    @arg("--schema-id", required=True, help="Schema ID")
+    def service__schema__get(self):
+        """Get Kafka Schema Registry schema"""
+        project_name = self.get_project()
+        schema = self.client.get_schema(project_name, self.args.name, self.args.schema_id)
+        self.print_response(schema)
+
+    @arg.project
+    @arg.service_name
+    @arg.subject
+    @arg.schema
+    def service__schema__create(self):
+        """Create Kafka Schema Registry schema"""
+        project_name = self.get_project()
+        schema = self.client.create_schema_subject_version(project_name, self.args.name, self.args.subject, self.args.schema)
+        self.print_response(schema)
+
+    @arg.project
+    @arg.service_name
+    @arg.subject
+    @arg.version_id
+    @arg.schema
+    def service__schema__check(self):
+        """Check Kafka Schema Registry schema compatibility"""
+        project_name = self.get_project()
+        schema = self.client.check_schema_compatibility(
+            project_name, self.args.name, self.args.subject, self.args.version_id, self.args.schema
+        )
+        self.print_response(schema)
+
+    @arg.project
+    @arg.service_name
+    def service__schema__configuration(self):
+        """Get Kafka Schema Registry global configuration"""
+        project_name = self.get_project()
+        self.print_response(self.client.get_schema_global_configuration(project_name, self.args.name))
+
+    @arg.project
+    @arg.service_name
+    @arg.compatibility
+    def service__schema__configuration_update(self):
+        """Update Kafka Schema Registry global configuration"""
+        project_name = self.get_project()
+        self.client.update_schema_global_configuration(project_name, self.args.name, self.args.compatibility)
+
+    @arg.project
+    @arg.service_name
+    @arg.subject
+    @arg.compatibility
+    def service__schema__subject_update(self):
+        """Update Kafka Schema Registry subject"""
+        project_name = self.get_project()
+        self.client.update_schema_subject_configuration(
+            project_name, self.args.name, self.args.subject, self.args.compatibility
+        )
+
+    @arg.project
+    @arg.service_name
+    @arg.subject
+    def service__schema__subject_configuration(self):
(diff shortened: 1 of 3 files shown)


aiven/client/cli.py
```diff
@@ -1298,6 +1298,147 @@ def service__connector__restart_task(self):
...
+        project_name = self.get_project()
+        self.print_response(self.client.get_schema_global_configuration(project_name, self.args.name))
+
+    @arg.project
+    @arg.service_name
+    @arg.compatibility
+    def service__schema__configuration_update(self):
+        """Update Kafka Schema Registry global configuration"""
+        project_name = self.get_project()
+        self.client.update_schema_global_configurationn(project_name, self.args.name, self.args.compatibility)
```
Ormod: configurationn -> configuration

aiven/client/cli.py
```diff
@@ -1298,6 +1298,147 @@ def service__connector__restart_task(self):
...
+
+    @arg.project
+    @arg.service_name
+    @arg.subject
+    @arg.version_id
+    @arg.schema
+    def service__schema__check(self):
+        """Check Kafka Schema Registry schema compatibility"""
+        project_name = self.get_project()
+        schema = self.client.check_schema_compatibility(project_name,
```
rikonen:
```
        schema = self.client.check_schema_compatibility(
            project_name,
            self.args.name,
            self.args.subject,
            self.args.version_id,
            self.args.schema,
        )
```

and
```
        schema = self.client.check_schema_compatibility(
            project_name, self.args.name, self.args.subject, self.args.version_id, self.args.schema
        )
```

are the options for splitting into multiple lines that should be used (even though this project doesn't have static checker to enforce that and the style isn't currently consistently applied everywhere). Same applies to other similar cases.

fmorato: ok. noted

aiven/client/cli.py
```diff
@@ -1298,6 +1298,141 @@ def service__connector__restart_task(self):
         project_name = self.get_project()
         self.client.restart_kafka_connector_task(project_name, self.args.name, self.args.connector, self.args.task)
 
+    @arg.project
+    @arg.service_name
+    @arg("schema-id", help="ID of a schema")
+    def service__schema__get(self):
+        """Get Kafka Schema Registry schema"""
+        project_name = self.get_project()
+        schema = self.client.get_schema(project_name, self.args.name, getattr(self.args, "schema-id"))
```
rikonen: No need to use `getattr` here. Just `self.args.schema_id` works.

aiven/client/cli.py
```diff
@@ -1298,6 +1298,141 @@ def service__connector__restart_task(self):
         project_name = self.get_project()
         self.client.restart_kafka_connector_task(project_name, self.args.name, self.args.connector, self.args.task)
 
+    @arg.project
+    @arg.service_name
+    @arg("schema-id", help="ID of a schema")
```
rikonen: `a schema` => `the schema` would sound better though general convention seems to be just `Schema ID`. Having multiple positional arguments is typically discouraged because remembering the order is difficult and using named parameters is preferred.

aiven/client/cliarg.py
```diff
@@ -61,3 +61,8 @@ def wrapped(self):
 arg.verbose = arg("-v", "--verbose", help="Verbose output", action="store_true", default=False)
 arg.connector_name = arg("connector", help="Connector name")
 arg.json_path_or_string = json_path_or_string
+arg.subject = arg("subject", help="Subject name")
+arg.version_id = arg("version_id", help="Subject version")
+arg.compatibility = arg("compatibility", choices=["BACKWARD", "FORWARD", "FULL", "NONE"],
+                        help="Choose a compatibility level for the subject")
+arg.schema = arg("schema", help="Schema string quote escaped")
```
rikonen: These should also preferably be named arguments instead of positional.

aiven/client/client.py
```diff
@@ -445,6 +445,55 @@ def restart_kafka_connector_task(self, project, service, connector_name, task_id
...
+        url = "/project/{}/service/{}/kafka/schema/subjects/{}"
+        return self.verify(self.delete, url.format(project, service, subject_name))
+
+    def get_schema_subject_version(self, project, service, subject_name, version):
+        url = "/project/{}/service/{}/kafka/schema/subjects/{}/versions/{}"
+        return self.verify(self.get, url.format(project, service, subject_name, version))
+
+    def get_schema_subject_version_schema(self, project, service, subject_name, version):
+        url = "/project/{}/service/{}/kafka/schema/subjects/{}/versions/{}/schema"
+        return self.verify(self.get, url.format(project, service, subject_name, version))
```
rikonen: Subject names can presumably also have unsafe characters and should be quoted. Also I noticed that a couple of the Kafka Connect related APIs don't do quoting for connector name. Maybe adding some method like

```
@staticmethod
def build_path(*parts):
    return "/" + "/".join(quote(part, safe="") for part in parts)
```

and always using that to build paths like

```
path = self.build_path("project", project, "service", service, "kafka", "schema", "subjects", subject_name, "versions", version, "schema)
```

would make sense.

aiven/client/cli.py
fmorato: Positional arguments with dashes doesn't seem to be converted to underscore. Only for named arguments. I'm converting `schema_id` to named argument, so I'll remove the `getattr`.

aiven/client/client.py
fmorato: Added `build_path` in PR #98.

PR #98, opened 2019-05-06
Refactor to always quote url parameters

fmorato:


aiven/client/client.py
```diff
@@ -678,15 +693,15 @@ def access_token_create(self, description, extend_when_used=False, max_age_secon
     def access_token_revoke(self, token_prefix):
         if any(c in token_prefix for c in "/+="):
             token_prefix = quote(token_prefix, safe="")
```
rikonen: Should remove this to prevent double quotation

aiven/client/client.py
```diff
@@ -678,15 +693,15 @@ def access_token_create(self, description, extend_when_used=False, max_age_secon
...
-        return self.verify(self.delete, "/access_token/{}".format(token_prefix))
+        return self.verify(self.delete, self.build_path("access_token", token_prefix))
 
     def access_token_update(self, token_prefix, description):
         if any(c in token_prefix for c in "/+="):
             token_prefix = quote(token_prefix, safe="")
```
rikonen: Same here, needs to be removed since quotation anyway happens later.

PR #108, opened 2019-06-27
Update README.rst

allen-munsch:
Adds a brief couple of lines on aiven-credentials.json and how to use it with an auth token.


Ormod:
Could you please squash your commits, we generally try to have as clean version history as possible.

allen-munsch:
@Ormod Thanks. Updated.

PR #111, opened 2019-07-16
add help command

mteiste:


scripts/avn
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env $PYTHON
+#! /usr/bin/env python3
```
ojarva: I think officially dropping python support should not be part of this PR.

ojarva: Also, when this is done, we should also drop tests for python2 from travis.

mteiste: fixed

PR #119, opened 2019-11-08
Allow creating/listing support tickets via client.

facetoe:


aiven/client/cli.py
```diff
@@ -2675,6 +2675,51 @@ def credits__claim(self):
...
+        """Create a support ticket"""
+        project_name = self.get_project()
+        ticket = self.client.create_ticket(
+            description=self.args.description,
+            project=project_name,
+            service=self.args.service,
+            severity=self.args.severity,
+            title=self.args.title,
+        )
+        self.print_response(result=ticket)
```
heikju: Should add `json=self.args.json` to print_response args to make `--json` flag work

aiven/client/cli.py
```diff
@@ -2675,6 +2675,51 @@ def credits__claim(self):
...
+            "title",
+            "project_name",
+            "service_name",
+            "create_time",
+            "description",
+            "update_time",
+            "user_email",
+            "user_real_name"
+        ]
+        self.print_response(result=tickets, table_layout=layout, json=False)
```
heikju: `json=self.args.json` here too if the command is to support raw json output with `--json`

aiven/client/cli.py
```diff
@@ -2675,6 +2675,51 @@ def credits__claim(self):
         if self.args.json:
             self.print_response(result, json=True)
 
+    @arg.json
+    @arg.project
+    @arg("--service", help="Related service name")
+    @arg("--severity", required=True, choices=["low", "high", "critical"], help="Ticket severity")
+    @arg("--title", required=True, help="Short description")
+    @arg("--description", required=True, help="Longer description")
+    def ticket_create(self):
```
ojarva: Even though a single underscore works, for consistency it is better to have double underscore for the first split as well (i.e., `ticket__create` and `ticket__list`). This is a very small thing, but makes searching for specific methods nicer, as you always know any breaks are with double underscore.

aiven/client/client.py
```diff
@@ -969,3 +969,17 @@ def claim_project_credit(self, project, credit_code):
...
+            "title": title,
+            "description": description,
+        }
+        if service:
+            body["service_name"] = service
+
+        return self.verify(self.post, self.build_path("project", project, "tickets"), body=body)
+
+    def list_tickets(self, project):
+        return self.verify(self.get, self.build_path('project', project, 'tickets'))
```
ojarva: Double quotes everywhere

PR #122, opened 2019-11-23
vpc: update user-defined peer network cidrs

mwfrojdman:
Supported in the CLI with the `vpc user-peer-network-cidr add` and
`delete` commands.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -1863,11 +1863,16 @@ ssl.truststore.type=JKS
             )
             raise argx.UserError(msg)
 
+    _peer_cloud_account_help = "AWS account ID, Google project ID, or Azure subscription ID"
+    _peer_resource_group_help = "Azure resource group name"
+    _peer_vpc_help = "AWS VPC ID, Google VPC network name, or Azure VNet name"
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -1995,6 +2000,47 @@ def _get_service_project_vpc_id(self):
...
+        """Add one ore more peer network CIDRs to a VPC"""
+        project_name = self.get_project()
+        add_base = {
+            "peer_cloud_account": self.args.peer_cloud_account,
+            "peer_vpc": self.args.peer_vpc,
+        }
+        if self.args.peer_resource_group is not None:
+            add_base["peer_resource_group"] = self.args.peer_resource_group
+        if self.args.peer_region is not None:
+            add_base["peer_region"] = self.args.peer_region
```
rikonen: This isn't listed as supported property in https://api.aiven.io/doc/#api-Project-VpcPeeringConnectionUpdate

mwfrojdman: Good catch. Removed. Should be added to the Aiven API, though currently it doesn't matter much as no `user_peer_network_cidr` compatible peering connection types support it.

PR #123, opened 2019-12-20
Migration status API and CLI

jappja:


aiven/client/cli.py
```diff
@@ -69,6 +69,8 @@ def convert_str_to_value(schema, str_value):
                                  .format(str_value, ", ".join(values)))
     elif "array" in schema["type"]:
         return [convert_str_to_value(schema["items"], val) for val in str_value.split(",")]
+    elif "null" in schema["type"]:
```
rikonen: Could also check that str_value is `None`, otherwise if we add new types later and forget to update this then the unsupported type gets converted into `None`, which is much harder to spot that getting explicit error.

aiven/client/cli.py
```diff
@@ -320,8 +322,18 @@ def collect_user_config_options(self, obj_def, prefix=""):
...
             full_name = prop if not prefix else (prefix + "." + prop)
-            if spec["type"] == "object":
+            types = spec["type"]
+            if not isinstance(types, list):
+                types = [types]
+            # "object" or ["object", "null"]
+            if "object" in types:
                 opts.update(self.collect_user_config_options(spec, prefix=full_name))
+                if "null" in types:
+                    # allow removing option with empty value
```
rikonen: Should we rather have some explicit switch for specifying that a value is to be deleted? It is possible that both null and empty string are valid values and making empty string mean that the value should be set to null can cause problems later on.

jappja: Yes I guess "service update -d <configuration> <service>" could work better.

aiven/client/cli.py
```diff
@@ -101,6 +103,82 @@ def add_args(self, parser):
...
+
+            conf[parts[-1]] = value
+
+        for opt in self.args.user_option_remove:
+            opt_schema = options.get(opt)
+            if not opt_schema:
+                raise argx.UserError("Unsupported option {!r}, available options: {}"
+                                     .format(opt, ", ".join(options) or "none"))
+
+            if opt_schema["type"] != "null":
```
rikonen: Shouldn't this rather do `if "null" not in opt_schema["type"]:` ? Current implementation only works for type object, not for e.g. string that allows both null and the string.

aiven/client/cli.py
```diff
@@ -101,6 +103,82 @@ def add_args(self, parser):
...
+                raise argx.UserError("Removing option {!r} is not supported".format(opt))
+
+            user_config[opt] = None
```
rikonen: Missing the logic for splitting first dot separated part so that `pg.foo.bar` gets set to structure like `user_config["pg"] = {"foo.bar": None}`

aiven/client/cli.py
```diff
@@ -560,13 +622,18 @@ def service__types(self):
...
                         description = ": {}".format(spec["description"]) if "description" in spec else ""
                         types = spec["type"]
-                        if not isinstance(types, list):
-                            types = [types]
-                        type_str = " or ".join(t for t in types if t != "null")
-                        print("  {title}{description}\n"
-                              "     => -c {name}=<{type}>  {default}"
-                              .format(name=name, type=type_str,
-                                      default=default_desc, title=spec["title"], description=description))
+                        if isinstance(types, str) and types == "null":
```
rikonen: Should first convert to list and then print the `--remove-option` option in case any of the list values is `"null"` so that this works also for data types other than object.

jappja: Then there's problem that all options accepting null as value will be reported for both add/remove.

rikonen: Shouldn't they be? Maybe the output format could be made more compact if it gets too cluttered. Like `(Use -d {name} to remove)` after the `-c {name}=<{type}>  {default}` part.

PR #141, opened 2020-07-08
Enable travis builds for multiple operating systems.

willrouesnel-aivenio:
Update `.travis.yml` to execute the Python test suite on all supported operating systems. Linter checks are configured to run as a separate stage and only on Linux since they should be OS-independent.

isort is version locked below the 5.x series since it is incompatible with the 5.x series it seems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,104 @@ python:
   - "3.8"
   - "pypy3"
 
-install:
-  - "pip install flake8 pylint pytest requests isort"
+os:
+  # Linux is the default platform for development.
+  - linux
 
+jobs:
+  include:
+    # Ensure we only run linters once.
+    - stage: lint
+      python: 3.8
+      os: linux
+      name: Style
+      script:
+        - pip install -r requirements.dev.txt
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,13 +6,105 @@ python:
...
+  - lint
+  - test
 
+# Don't use the Makefile since it won't work on Windows without a lot of effort.
 script:
-  - "make flake8"
-  - "make pylint"
-  - "make pytest"
-  - "make validate-style"
+  - $PYTHON -m pytest -vv tests/
```
rominf, line 110: Could you add a newline? :)

.travis.yml
```diff
@@ -6,13 +6,105 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
-  - "pypy3"
+
```
rominf, line 9: Why did you drop PyPy?

willrouesnel-aivenio: AFAIK I don't think we call it out in the readme/metadata that we do support it, but you're right, it should be left in. Fixed.

PR #143, opened 2020-07-08
Fix CI by using an old version of isort

rominf:
See https://github.com/timothycrosley/isort/issues/1273


rominf:
Closed in favor of #141

PR #146, opened 2020-07-13
Add uniform support for data output format.

willrouesnel-aivenio:
Modifies argx to collect a uniform set of specifiers at launch for log level, output format and output stream. This is passed to all command-line utilities and affects output via the `print_response` function.

This centralizes output selection and enables all commands to emit output using any supported format. YAML via the `ruamel.yaml` library is added as a supported format as a part of this patch.

`print_response` is shimmed to allow this version of the library to be backwards compatible, pending removal of redundant parameters from other tools.

Cross-tested as backwards compatible (commands work, output looks sensible) with the existing dependent tooling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ jobs:
       name: PyLint
       script:
         - pip install -r requirements.dev.txt
+        - pip install -e .
         - make pylint
 
     # MacOS we want to ensure system-python is working properly
--- a/aiven/client/argx.py
+++ b/aiven/client/argx.py
@@ -2,18 +2,25 @@
 #
 # This file is under the Apache License, Version 2.0.
 # See the file `LICENSE` for details.
-
 from aiven.client import envdefault, pretty
+from collections.abc import Mapping, Sequence
 
 import aiven.client.client
 import argparse
 import csv as csvlib
+import enum
 import errno
 import json as jsonlib
 import logging
 import os
 import requests.exceptions
+import ruamel.yaml as yamllib
 import sys
+
+try:
+    basestring
+except NameError:
+    basestring = str  # pylint: disable=redefined-builtin
 
 # Optional shell completions
 try:
@@ -99,19 +106,76 @@ class Config(dict):
             jsonlib.dump(self, fp, sort_keys=True, indent=4)
 
 
-class CommandLineTool:  # pylint: disable=old-style-class
+class OutputFormats(enum.Enum):
+    """Defines the supported output formats on the command line."""
+    TABLE = "table"
+    TABLE_NOHEADER = "table-noheader"
+    CSV = "csv"
+    CSV_NOHEADER = "csv-noheader"
+    JSON = "json"
+    JSON_COMPACT = "json-compact"
+    YAML = "yaml"
+
+    def __str__(self):
+        return str(self.value)
+
+
+class CommandLineTool:  # pylint: disable=old-style-class,too-many-instance-attributes
     def __init__(self, name):
         self.log = logging.getLogger(name)
         self.config = None
         self._cats = {}
         self._extensions = []
-        self.parser = argparse.ArgumentParser(prog=name, formatter_class=CustomFormatter)
-        self.parser.add_argument(
-            "--config",
-            help="config file location %(default)r",
-            default=envdefault.AIVEN_CLIENT_CONFIG,
+
+        if name is None:
+            raise ValueError("name cannot be None")
+
+        if not isinstance(name, basestring):
+            raise ValueError("name must be a string")
+
+        self._output_format = OutputFormats.TABLE  # type: OutputFormats
+        self._output_stream = sys.stdout
+        self._single_item = False
+
+        # Early parser handles things like logging which need to be setup as early as possible.
+        # Output from early parser is never shown.
+        self.early_parser = argparse.ArgumentParser(
+            prog=name, formatter_class=CustomFormatter, add_help=False
+        )  # Important - help goes to real parser
+        self.early_parser.add_argument(
+            "--log-level",
+            choices=list(logging._nameToLevel),  # pylint: disable=protected-access
+            default=logging._levelToName[logging.INFO],
+            help="Log level"
         )
-        self.parser.add_argument("--version", action="version", version="aiven-client {}".format(__version__))
+        self.early_parser.add_argument(
+            "--request-log-level",
+            choices=list(logging._nameToLevel),
+            default=logging._levelToName[logging.WARNING],
+            help="HTTP request log level"
+        )
+        self.early_parser.add_argument(
+            "--output-format",
+            type=OutputFormats,
+            choices=list(OutputFormats),
+            default=self._output_format,
+            help="Output format"
+        )
+        self.early_parser.add_argument("--output", type=str, default="-", help="Output destination")
+
+        self.early_parser.add_argument(
+            "--config", help="config file location %(default)r", default=envdefault.AIVEN_CLIENT_CONFIG
+        )
+
+        # Parser is the actual command line parser which displays to the user.
+        self.parser = argparse.ArgumentParser(prog=name, formatter_class=CustomFormatter)
+        self.parser.add_argument('--version', action='version', version='aiven-client {}'.format(__version__))
+
+        # Add the early parser options to the real parser so they'll appear in the help we show the
+        # user.
+        for action in self.early_parser._actions:
+            self.parser._add_action(action)
+
         self.subparsers = self.parser.add_subparsers(title="command categories", dest="command", help="", metavar="")
         self.args = None
 
@@ -148,13 +212,7 @@ class CommandLineTool:  # pylint: disable=old-style-class
             parser.add_argument(*arg_prop[0], **arg_prop[1])
 
     def add_args(self, parser):
-        pass  # override in sub-class
-
-    def extend_commands(self, sub_client):
-        """Add top-level args and all commands from a CommandLineTool instance"""
-        sub_client.add_args(self.parser)  # top-level args
-        sub_client.add_cmds(self.add_cmd)  # sub-commands
-        self._extensions.append(sub_client)
+        """This method should be overriden in subclasses to add group arguments."""
 
     def add_cmds(self, add_func):
         """Add every method tagged with @arg as a command"""
@@ -166,6 +224,12 @@ class CommandLineTool:  # pylint: disable=old-style-class
             func = getattr(self, prop, None)
             if getattr(func, ARG_LIST_PROP, None) is not None:
                 add_func(func)
+
+    def extend_commands(self, sub_client):
+        """Add top-level args and all commands from a CommandLineTool instance"""
+        sub_client.add_args(self.parser)  # top-level args
+        sub_client.add_cmds(self.add_cmd)  # sub-commands
+        self._extensions.append(sub_client)
 
     def parse_args(self, args=None):
         self.extend_commands(self)
@@ -186,7 +250,7 @@ class CommandLineTool:  # pylint: disable=old-style-class
     def print_response(
         self,
         result,
-        json=True,
+        json=False,
         format=None,  # pylint: disable=redefined-builtin
         drop_fields=None,
         table_layout=None,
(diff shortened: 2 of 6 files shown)


aiven/client/argx.py
```diff
@@ -186,7 +250,7 @@ def expected_errors(self):
     def print_response(
         self,
         result,
-        json=True,
+        json=False,
```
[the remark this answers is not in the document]
willrouesnel-aivenio: The basic problem is this default is kind of silly, because AFAIK everywhere else this is used `json` is overridden by `args.json` which is a command line flag that evals to `False`. So conventionally, anywhere in our code we a have an explicit `--json` flag on the command, this is actually defaulting to false.

From a UI consistency perspective this isn't great: if you add a new command and use `print_response` naively, unless you stick an argx.json flag on there, you get JSON by default rather then pretty table everywhere there _is_ a `--json` option.

I suppose to backwards compatibility this, `json=None` could be specified, and we detect that as `json=True` (but unspecified) which means the code can detect if a default was specified and then choose whether to use the global format selection if _that_ was specified.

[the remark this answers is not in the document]
facetoe: Agree it is a silly default. I guess the main thing is we don't change the output accidentally and break any user code/scripts. If you can be sure that this change won't have that effect then it should be OK.

.travis.yml
```diff
@@ -37,6 +37,7 @@ jobs:
       name: PyLint
       script:
         - pip install -r requirements.dev.txt
+        - pip install -e .
```
facetoe, line 40: Curious why this was added and when it wasn't necessary previously.

aiven/client/argx.py
```diff
@@ -196,47 +260,100 @@ def print_response(
...
         """print request response in chosen format"""
-        if file is None:
-            file = sys.stdout
+
+        # Deprecation compatibility: map the old keyword args into the new enum:
+        output_stream = self._output_stream if file is None else file
+
+        output_format = self._output_format
+
+        if json is True:
```
facetoe, line 269: Is there a reason to check for `True` explicitly?

aiven/client/argx.py
```diff
@@ -196,47 +260,100 @@ def print_response(
...
+            jsonlib.dump(result, output_stream, indent=4, sort_keys=True, cls=pretty.CustomJsonEncoder)
+            output_stream.write("\n")
+        elif output_format == OutputFormats.JSON_COMPACT:
+            jsonlib.dump(result, output_stream, sort_keys=True, cls=pretty.CustomJsonEncoder)
+            output_stream.write("\n")
+        elif output_format == OutputFormats.YAML:
+            yamllib.round_trip_dump(result, output_stream, default_flow_style=False)
+
+        else:
+            raise NotImplementedError(
```
facetoe, line 326: ValueError would be more appropriate IMO - https://docs.python.org/3/library/exceptions.html

aiven/client/argx.py
```diff
@@ -196,47 +260,100 @@ def print_response(
...
+        )
+        logging.getLogger("requests").setLevel(
+            logging._nameToLevel[early_args.request_log_level]  # pylint: disable=protected-access
+        )
+
+        self.config = Config(early_args.config)
+
+        self._output_format = early_args.output_format
+
+        if early_args.output == "-" or early_args.output == "":
```
facetoe, line 351: If `-` means stdout then might as well just call it stdout. I don't really see a reason to use a single dash to represent it unless I am missing something.

aiven/client/pretty.py
```diff
@@ -62,69 +89,127 @@ def format_item(key, value):
...
+class PreparedTableData:
+    """PreparedTableData represents prepared data for methods which render tables."""
+
+    def __init__(self, fields, table_rows, horizontal_fields, vertical_fields):
+        self._fields = fields
+        self._table_rows = table_rows
+        self._horizontal_fields = horizontal_fields
+        self._vertical_fields = vertical_fields
+
+    @property
```
facetoe, line 101: This could be a simple class to hold the values, the use of @property doesn't seem to add any benefit from accessing the attributes directly.

aiven/client/pretty.py
```diff
@@ -62,69 +89,127 @@ def format_item(key, value):
...
+    def vertical_fields(self):
+        return self._vertical_fields
+
+
+def prepare_table(result, table_layout=None):
+    if result is None:
+        return PreparedTableData([], [], [], [])
+
+    if not isinstance(result, (Sequence, )) or isinstance(result, (basestring, )):
```
facetoe, line 122: There are many of these sorts of checks, might be good to refactor them out into some helper methods. Also not sure why a tuple containing one item is being passed to `isinstance`.

.travis.yml
willrouesnel-aivenio: It's because of the dependency on a yaml library. PyLint wants to import the code, so it needs the import to resolve. Since we now have 1 extra 3rd party dependency, we need to do the install for online linting to pass.

aiven/client/argx.py
willrouesnel-aivenio: `json` isn't just "kind of truthy" it should distinctly be True or False. This encourages test suites to fail if some undefined behavior is being used (better would be for mypy to just catch it but we're not there yet).

aiven/client/argx.py
willrouesnel-aivenio: Disagree on this one: the entry point to this function via the enum means this code should never be reached. If you do reach it, it's because you extended the enum without adding the handler.

aiven/client/argx.py
willrouesnel-aivenio: Unix command line convention: `-` is the common specifier for stdout, whereas if you enter anything else then it's treated as a filename/filepath.

aiven/client/pretty.py
willrouesnel-aivenio: It's a little bit of future proofing: all the table routines need 4 basic bits of data - a list of fields, a list of rows, and lists of which fields are horizontal and vertical.

Currently this calculation happens in `prepare_table` for want of a clear idea of where it's going. Possibly the object in the future on initialization should instead just receive the data and set that up internally.

.travis.yml
facetoe: This is the sort of information that would be handy in a commit message.

aiven/client/pretty.py
willrouesnel-aivenio: AFAIK (it may have changed in more recent versions) `isinstance` has always wanted a tuple.

aiven/client/argx.py
facetoe: I'd say that the enum passed in is a value that cannot be handled, therefore ValueError. I guess you could argue either point, however I don't think this case matches the intent for NotImplementedError:

>  In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

aiven/client/argx.py
facetoe: This is true, but it doesn't mean that we need to adopt it here. I'd argue that `-` is less descriptive than `stdout` so might as well go with the latter.

aiven/client/pretty.py
facetoe: If at some later date it needs to be modified it is easy to do without breaking any of the code that depends on it. As it stands currently there doesn't seem to be any reason to use a property.

aiven/client/pretty.py
facetoe: Yep fair enough. It is not necessary to pass a tuple though so might as well just pass the type.

aiven/client/argx.py
```diff
@@ -196,47 +260,100 @@ def print_response(
...
+
+        output_format = self._output_format
+
+        if json is True:
+            output_format = OutputFormats.JSON
+        elif csv is True:
+            output_format = OutputFormats.CSV if header is True else OutputFormats.CSV_NOHEADER
+
+        # Detect when a single item has been passed and reframe it as a list.
+        if single_item is True or not isinstance(result, (Sequence, )) or isinstance(result, (str, )):
```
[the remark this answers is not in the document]
willrouesnel-aivenio: The problem with iterable is that a mapping is iterable, but existing behavior requires that if we get a mapping it's treated as a single item.

aiven/client/argx.py
willrouesnel-aivenio: The problem is `stdout` is reasonably a file path, not stdout. A user at their console can actually ask bash to pass `/dev/stdout` if they must write literally what they want, but otherwise most common unix commands use this convention - 
```
$ curl -o - google.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="http://www.google.com/">here</A>.
</BODY></HTML>

$ wget -O - google.com
# truncated since it follows links
```

aiven/client/argx.py
```diff
@@ -99,19 +106,76 @@ def save(self):
...
         self._cats = {}
         self._extensions = []
-        self.parser = argparse.ArgumentParser(prog=name, formatter_class=CustomFormatter)
-        self.parser.add_argument(
-            "--config",
-            help="config file location %(default)r",
-            default=envdefault.AIVEN_CLIENT_CONFIG,
+
+        if name is None:
+            raise Exception("name cannot be None")
```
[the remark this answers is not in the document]
willrouesnel-aivenio: Fixed.

aiven/client/argx.py
```diff
@@ -99,19 +106,76 @@ def save(self):
...
+
+        if not isinstance(name, (str, )):
+            raise Exception("name must be a string")
```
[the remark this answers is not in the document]
willrouesnel-aivenio: Fixed.

aiven/client/argx.py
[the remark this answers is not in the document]
willrouesnel-aivenio: Need a call on this one then. A non-thorough (i.e. programmatic) look through our existing interface suggests it would change nothing about the commands in `aiven-client`.

.travis.yml
willrouesnel-aivenio: Fixed.

aiven/client/argx.py
willrouesnel-aivenio: Fixed.

aiven/client/pretty.py
willrouesnel-aivenio: Fixed.

aiven/client/pretty.py
willrouesnel-aivenio: Fixed.

aiven/client/argx.py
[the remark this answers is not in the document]
lionbee: @ojarva do you have an opinion here?

tests/test_argx.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2015, Aiven, https://aiven.io/
```
rominf, line 1: Old copyright year.

tests/test_argx.py
```diff
@@ -0,0 +1,97 @@
...
+            "inconsistent": "keys"
+        }],
+    ]
+)
+@pytest.mark.parametrize("fmt", list(OutputFormats))
+def test_print_response(request, value, fmt):
+    cli = CommandLineTool(request.node.name)
+    cli._output_format = fmt  # pylint: disable=protected-access
+    print("\n")
+    cli.print_response(value)
```
rominf, line 56: This test doesn't check the resulting output. I think checking the result is necessary for such big patches like this one.

tests/test_argx.py
```diff
@@ -0,0 +1,97 @@
...
+        [("nested.a", "nested.b"), "something_else"],
+    ]
+)
+@pytest.mark.parametrize(
+    "fmt", [OutputFormats.TABLE, OutputFormats.TABLE_NOHEADER, OutputFormats.CSV, OutputFormats.CSV_NOHEADER]
+)
+def test_print_response_with_vertical_fields(request, value, fmt, table_layout):
+    cli = CommandLineTool(request.node.name)
+    cli._output_format = fmt  # pylint: disable=protected-access
+    cli.print_response(value, table_layout=table_layout)
```
rominf, line 96: Same as above.

facetoe:
Closing this as there has been no activity in over a month.

PR #147, opened 2020-08-24
cli: fix unbound variables, formatting

joukost:
Subcommands to update integration or integration endpoint fail with an `UnboundLocalError` unless config variables are also supplied.

Also removed redundant parentheses.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -56,8 +56,10 @@ def convert_str_to_value(schema, str_value):
         }
         try:
             return values[str_value]
-        except KeyError:
-            raise argx.UserError("Invalid boolean value {!r}: expected one of {}".format(str_value, ", ".join(values)))
+        except KeyError as ex:
+            raise argx.UserError(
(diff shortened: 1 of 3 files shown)


aiven/client/cli.py
```diff
@@ -1317,9 +1317,9 @@ def service__integration_endpoint_create(self):
     @arg.json
     def service__integration_endpoint_update(self):
         """Update a service integration endpoint"""
+        endpoint_id = getattr(self.args, "endpoint-id")
```
rikonen, line 1320: Fwiw I think it would be better to name the argument `endpoint_id` to avoid the use of `getattr()`.

rikonen (approved):

rikonen (changes requested):
Looks like pylint version has changed at this is complaining about some existing issue

aiven/client/cli.py
joukost: Makes sense, it's positional anyway. I'll change it...

joukost:
Apparently my local `yapf` didn't agree too well either.

rikonen (approved):

PR #149, opened 2020-10-12
Documentation for decorator arg

hackaugusto:
This adds a bit of documentation to `arg`, the intention is to make it a bit easier to understand it's usage.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -31,9 +31,7 @@ AUTHENTICATION_METHOD_COLUMNS = [
 PLUGINS = []
 
 try:
-    from aiven.admin import (
-        plugin as adminplugin,
-    )  # pylint: disable=import-error,no-name-in-module
+    from aiven.admin import plugin as adminplugin  # pylint: disable=import-error,no-name-in-module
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -31,9 +31,7 @@
 PLUGINS = []
 
 try:
-    from aiven.admin import (
-        plugin as adminplugin,
-    )  # pylint: disable=import-error,no-name-in-module
+    from aiven.admin import plugin as adminplugin  # pylint: disable=import-error,no-name-in-module
```
rominf, line 34: Unrelated change?

hackaugusto: My setup runs `yapf` automatically. Should I revert it?

rominf: Well, I think it's not that important, moreover, if it's a `yapf` related change. I'll merge this as is :)

rominf (approved):

aiven/client/cli.py
hackaugusto: thanks!

PR #154, opened 2020-10-28
argx: Sort subcommands alphabetically

willrouesnel-aivenio:
When using argx to build multiple tools, the subcommands weren't being
kept sorted. This patch ensures that the ordering of subcommands is
always alphabetical as opposed to based on the order of loading (which
is not consistent).


rikonen (dismissed):
Static checks fail

facetoe (approved):

PR #159, opened 2020-11-13
cli: self describing name

hackaugusto:
`name` is quite broad and sometimes it is hard to tell what name is necessary from the context, examples:

```
avn service keypair get --key-filepath /tmp/keyfile --cert-filepath /tmp/certfile
usage: avn service keypair get [-h] [--project PROJECT] --key-filepath
                               KEY_FILEPATH --cert-filepath CERT_FILEPATH
                               name keypair
avn service keypair get: error: the following arguments are required: name, keypair
```
and
```
avn service ca get --target-filepath /tmp/test.ca
usage: avn service ca get [-h] [--project PROJECT] --target-filepath
                          TARGET_FILEPATH
                          name ca
avn service ca get: error: the following arguments are required: name, ca
```

Calling the argument `service_name` should make it slightly easier to guess what value should be used.


fingon (approved):

mts-aiven:
Breaks, functions in cli.py expect the value in self.args.name and not self.args.service_name

PR #163, opened 2020-11-18
cli: Fix service_name usage

mts-aiven:
Was broken in #159 / 51b30e6 (change halfway)

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -394,7 +394,7 @@ class AivenCLI(argx.CommandLineTool):
                     project=self.get_project(),
                     limit=self.args.limit,
                     offset=previous_offset,
-                    service=self.args.name,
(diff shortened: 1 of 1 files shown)


hackaugusto:
My bad, I assumed the tests would fail if the change was broken.

Btw, AFAIK `arg` is just a proxy to [`add_argument`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument), so another solution is just to do a:

```diff
- arg.service_name = arg("service_name", help="Service name")
+ arg.service_name = arg("service_name", help="Service name", dest="name`)
```

mts-aiven:
Thought about that but 'name' is also used in different ways in cli.py; also using same argument name makes cli.py more readable in my opinion.

hackaugusto (approved):
LGTM, thanks!

PR #164, opened 2020-11-18
bugfix: revert service_name to name

hackaugusto:
The name of the variable should be kept `name` to not break existing
code, instead only the user facing string should be changed.

PR https://github.com/aiven/aiven-client/pull/159 broke the cli because it also changed the name of the variable used in the namespace `AttributeError: 'Namespace' object has no attribute 'name'`. This maintains the name as `name` and only changes the text used for the help text. [Referece](https://docs.python.org/3/library/argparse.html#metavar)


hackaugusto:
closing in favor of https://github.com/aiven/aiven-client/pull/163

PR #165, opened 2020-11-20
cli: fix some `service` and `project` subcommands

heikju:
3d2b577 either shouldn't have touched some of these, or the arg rename was incomplete.


ilpon (changes requested):
Also `service terminate`, `project delete` and `project update` are broken.

ilpon (approved):

PR #166, opened 2020-11-24
Fix some account commands.

heikju:
Revert some more of 3d2b5771, which did a bit too much of search & replacing.


mts-aiven (approved):
LGTM

PR #168, opened 2020-11-30
Deprecate / remove service group concept

rikonen:
The service group concept has been removed on backend side. The API
still accepts and returns the related fields but the input is always
ignored and output is hardcoded dummy value. Remove it also from the
cli except for service create and update commands so that backwards
compatibility is kept.

Note that the Python library change is backwards incompatible as it
was not possible to make the parameter optional while retaining
compatibility.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -816,7 +816,6 @@ class AivenCLI(argx.CommandLineTool):
         "state",
         "cloud_name",
         "plan",
-        "group_list",
         "create_time",
         "update_time",
     ]]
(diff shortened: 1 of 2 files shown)


mwfrojdman (dismissed):
It could make sense using `warnings.warn` with `DeprecationWarning`. See other cases in the same repository.

rikonen:
`warnings.warn("whatever", DeprecationWarning)` for whatever reason doesn't print out anything at all. Looking at the source code it should print to sys.stderr but it doesn't (manually printing to sys.stderr does yield output both before and after the `warnings.warn()` call).

mwfrojdman:
Now that you mention it, IIRC the library module was the place to `warnings.warn`. The CLI probably should just print a warning with logging like you did

mwfrojdman (approved):

PR #170, opened 2020-12-15
cli: fix format for single item

pintor:
Currently, `---format` fails for `service get` with the following Exception:

``TypeError: format() argument after ** must be a mapping, not str``

This change fixes `--format` for all calls to `print_response` when `single_item` is `True`

EDIT:
Added fix for csv output as well


rikonen:
Since you're touching this maybe fix `csv` too? It looks to have the same issue.

rikonen (approved):

PR #172, opened 2020-12-29
security: add security policy

tapioaiven:
Add security policy with advise on reporting possible vulnerabilities.


Myrcurial (approved):
Approved. No changes needed. Repeat across public repos.

PR #175, opened 2021-01-21
Add service user get and set-access-control

ilpon:


encima (commented):
Works well and definitely useful for those not using TF, thanks for this!

As it is adding new functionality, is it worth adding a test case?

ilpon:
> As it is adding new functionality, is it worth adding a test case?

As there are no existing tests like that, it would require some thinking how the tests should work here. This is being used internally by Aiven so it is being tested outside of the repository.

aiven/client/cli.py
```diff
@@ -1209,9 +1209,8 @@ def service__user_create(self):
         arg_vars = vars(self.args)
         extra_params = {}
         acl_params = {
-            key: arg_vars[key].split()
+            key: arg_vars[key].split() if arg_vars[key] else []
```
heikju, line 1212: Does this mean that were the user to give e.g. just `--redis-acl-keys`, the command and category lists would be emptied? If so, that leaves some unnecessary room for user error IMO, and it would be safer to require all of the three args be given, if any of them is.

heikju (dismissed):

aiven/client/cli.py
ilpon: As this is a generic command and the arguments work only for Redis, it would be a bit complicated to require them.
I don't see this as a big problem as the defaults are safe and there is now a separate command to modify these. The logic also matches quite closely the logic in Redis.

ilpon: The `service user-set-access-control` changes only the keys, if only keys are given. Same goes for commands and categories. The command would be quite awkward to use, if you would have to give all of those every time.

I was more worried that changing commands would clear the categories or changing the categories would clear the commands as those are actually defining the same thing, but it does actually work so that setting only commands won't affect the set categories and setting only categories won't affect the set commands. 

The user creation could require all three, but I still don't see the need for that either and it isn't possible to just mark the arguments required so it would need to be a separate check that affects only Redis.

ilpon: Updated the create and update to share the parsing logic. Now when option is not given, it will be passed as None and an empty string is passed as empty array. For creation this does not currently make difference, but if we were to change the defaults, it would. For the update it makes it change only the given values.

heikju: Thanks, I believe only changing on update what's explicitly requested indeed makes more sense.

PR #178, opened 2021-01-22
Replace travis with GitHub actions

lornajane:
Travis is no longer building since the service is discontinued. Replace the CI on this project with GitHub Actions.

(note that we probably want to edit the repo settings to require these tests to pass before merging once we accept this into master)


rominf (approved):
LGTM, but I'll let to check this to someone else, as I'm not an expert in GitHub Actions.

rikonen (approved):

PR #180, opened 2021-01-27
Add python 3.9 to the build matrix

lornajane:
We moved the builds to GitHub Actions, but we didn't update to include the current version of Python, 3.9. This is now included.


rominf (approved):

rominf:
@lornajane Thanks!
How about updating the classifiers in `setup.py` in a follow-up PR?

PR #182, opened 2021-01-28
gh_actions: only trigger in pull requests/tags and add filter for push

NARSimoes:
Currently we trigger the workflow file for every push and pull request event. For example, every time we create a PR we run 2x, when we create a branch we run, etc.

Is this the goal? Or shoud we start using the following conditions?
- `push`: trigger only when we merge to master
- `tags`: every time we create a tag
- `pull request`: every time we create and update a pull request

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,6 +1,12 @@
 name: Build Aiven Client
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


lornajane (approved):
This looks good, it will remove duplicate checks on the pull requests from branches on this repo - it does mean we won't get checks on those branches until we open the PR but that seems OK to me.

PR #184, opened 2021-03-08
Add "did you mean" hint on wrong option selection

rominf:
I used the code of Norvig, as it's small, and I didn't want to introduce additional dependency.

### DIFF
--- a/aiven/client/speller.py
+++ b/aiven/client/speller.py
@@ -1,0 +1,35 @@
+# Spelling Corrector in Python 3; see http://norvig.com/spell-correct.html
+#
+# Copyright (c) 2007-2016 Peter Norvig
+# MIT license: www.opensource.org/licenses/mit-license.php
+
+
+def suggest(word_to_check, known_words):
+    def get_correction(word):
+        """Most probable spelling correction for word."""
+        candidates = get_candidates(word)
+        return next(iter(candidates)) if candidates else None
+
+    def get_candidates(word):
+        """Generate possible spelling corrections for word."""
+        return get_known([word]) or get_known(get_edits1(word)) or get_known(get_edits2(word)) or []
+
+    def get_known(words):
+        """The subset of `words` that appear in the dictionary of WORDS."""
+        return set(w for w in words if w in known_words)
+
+    def get_edits1(word):
+        """All edits that are one edit away from `word`."""
+        letters = "abcdefghijklmnopqrstuvwxyz_"
+        splits = [(word[:i], word[i:]) for i in range(len(word) + 1)]
+        deletes = [L + R[1:] for L, R in splits if R]
+        transposes = [L + R[1] + R[0] + R[2:] for L, R in splits if len(R) > 1]
+        replaces = [L + c + R[1:] for L, R in splits if R for c in letters]
+        inserts = [L + c + R for L, R in splits for c in letters]
+        return set(deletes + transposes + replaces + inserts)
+
+    def get_edits2(word):
+        """All edits that are two edits away from `word`."""
+        return (e2 for e1 in get_edits1(word) for e2 in get_edits1(e1))
+
+    return get_correction(word_to_check)
--- a/tests/test_speller.py
+++ b/tests/test_speller.py
@@ -1,0 +1,40 @@
+# Copyright 2021, Aiven, https://aiven.io/
+#
+# This file is under the Apache License, Version 2.0.
+# See the file `LICENSE` for details.
+from aiven.client.speller import suggest
+
+import pytest
+
+# Current service types (2021-03-09); they are used just for testing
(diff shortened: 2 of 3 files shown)


aiven/client/speller.py
```diff
@@ -0,0 +1,40 @@
...
+        """Generate possible spelling corrections for word."""
+        return get_known([word]) or get_known(get_edits1(word)) or get_known(get_edits2(word)) or []
+
+    def get_known(words):
+        """The subset of `words` that appear in the dictionary of WORDS."""
+        return set(w for w in words if w in word_dict)
+
+    def get_edits1(word):
+        """All edits that are one edit away from `word`."""
+        letters = 'abcdefghijklmnopqrstuvwxyz'
```
pavdmyt, line 28: What happens if `word` contains capital letters?

aiven/client/speller.py
```diff
@@ -0,0 +1,40 @@
...
+# This file is under the Apache License, Version 2.0.
+# See the file `LICENSE` for details.
+
+# Spelling Corrector in Python 3; see http://norvig.com/spell-correct.html
+#
+# Copyright (c) 2007-2016 Peter Norvig
+# MIT license: www.opensource.org/licenses/mit-license.php
+
+
+def suggest(word_to_check, word_dict):
```
pavdmyt, line 12: There should be a 3rd party lib with similar functionality implemented. The benefit of such lib is that it is up to date and maintained as opposed to blogpost code.

rominf: Please propose a stable lightweight (without dictionaries included) library, as I didn't find any. Also, it's possible that we will start using a similar approach on the server-side, hence it would be great to have a Fedora RPM for that library.

aiven/client/speller.py
rominf: It works, you can check it :-)

aiven/client/speller.py
```diff
@@ -0,0 +1,40 @@
...
+    def get_correction(word):
```
nkchern-avn, line 13: would it make sense to extract these small helper functions out of `suggest`, so that it will be easier to observe what does `suggest` exactly do?

aiven/client/speller.py
lornajane: It doesn't quite do what I expected though, if I try with `-t kaaka` it correctly suggests "kafka" but if I do `-t KAAKA` then it doesn't prompt me with "did you mean" ... should it? I assume none of the other arguments are uppercase so I'm not sure if we need to support this.

lornajane (approved):
This is a nice touch, and an improvement on the current state that caught me out recently! Thanks for making this patch, it's a big APPROVE from me, but let's also address or resolve the other comments before we merge.

aiven/client/speller.py
rominf: Well, while it's possible to check this, I don't think we should. Google also supports the wrong keyboard layout (when I type "test" in Russian layout I get "еуые" -> Google proposes "test"). While it's possible to support this mistake too, it's too much of a wasted effort in my opinion.

aiven/client/speller.py
rominf: I don't agree with this. I think that this function is quite small and is good as is.

rominf:
Converted to draft, as there could be issues with dual licensing. Waiting for @lornajane to help me address this.

aiven/client/speller.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2021, Aiven, https://aiven.io/
+#
+# This file is under the Apache License, Version 2.0.
+# See the file `LICENSE` for details.
```
lornajane, line 4: I'm not a lawyer but I think if this file is under an MIT license, then that should be the only license in the file. It's fine that the code was also modified but it needs to retain its license. I don't think we can dual-license like this, so remove this section.

tests/test_speller.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2015, Aiven, https://aiven.io/
```
lornajane, line 1: Copyright 2021, please

aiven/client/speller.py
rominf: Done!

rominf:
@lornajane, I addressed your comments; I think it's ready for merge now.

aiven/client/speller.py
rominf: Or should have I keep the first line (Aiven copyright)?

rominf:
I'll have a talk with our legal team today. Moving to draft for now.

rominf:
If I understood @saaros correctly, this is now correct.

lornajane (approved):
Works well, thanks for checking the licensing side as well

PR #185, opened 2021-03-18
Fix CI badge point to GitHub Actions instead of Travis CI

tvainika:
Some leftovers from #178


lornajane (approved):
Good spot! Thanks for fixing this.

PR #186, opened 2021-03-19
cli: add support for topic tags

tvainika:
Add support for adding tags to topic while creating topics and adding, removing and updating tags while updating topics.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -6,6 +6,7 @@ from . import argx, client
 from aiven.client import envdefault
 from aiven.client.cliarg import arg
 from aiven.client.speller import suggest
+from collections import Counter
 from decimal import Decimal
 from urllib.parse import urlparse
 
@@ -65,6 +66,31 @@ def convert_str_to_value(schema, str_value):
         return None
     else:
         raise argx.UserError("Support for option value type(s) {!r} not implemented".format(schema["type"]))
+
+
+tag_key_re = re.compile(r"[\w\-]+")
+tag_value_re = re.compile(r"[\w\-,. ]*")
+
+
+def parse_tag_str(kv):
+    k, v = (kv.split(sep='=', maxsplit=1) + [''])[:2]
+
+    if not tag_key_re.fullmatch(k):
+        raise argx.UserError(f"Tag key '{k}' must consist of alpha-numeric characters, underscores or dashes")
+
+    if not tag_value_re.fullmatch(v):
+        raise argx.UserError(
+            f"Tag value '{k}={v}' must consist of alpha-numeric characters, underscores, dashes, commas or dots"
+        )
+
+    return {"key": k, "value": v}
+
+
+def parse_untag_str(k):
+    if not tag_key_re.match(k):
+        raise argx.UserError(f"Tag key {k} must consist of alpha-numeric characters, underscores or dashes")
+
+    return k
 
 
 def no_auth(fun):
@@ -1914,6 +1940,8 @@ ssl.truststore.type=JKS
         for topic in topics:
             if topic["retention_hours"] == -1:
                 topic["retention_hours"] = "unlimited"
+            if not self.args.json:
+                topic["tags"] = [f"{t['key']}={t['value']}" for t in topic["tags"]]
         layout = [[
             "topic_name",
             "partitions",
@@ -1922,6 +1950,7 @@ ssl.truststore.type=JKS
             "retention_bytes",
             "retention_hours",
             "cleanup_policy",
+            "tags",
         ]]
         self.print_response(topics, format=self.args.format, json=self.args.json, table_layout=layout)
 
@@ -2012,6 +2041,7 @@ ssl.truststore.type=JKS
     @arg.min_insync_replicas
     @arg.retention
     @arg.retention_bytes
+    @arg.tag
     @arg(
         "--cleanup-policy",
         help="Topic cleanup policy",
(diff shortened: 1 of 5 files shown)


aiven/client/cli.py
```diff
@@ -2020,6 +2049,14 @@ def service__task_create(self):
     )
     def service__topic_create(self):
         """Create a Kafka topic"""
+
+        tags = list(map(parse_tag_str, self.args.topic_option_tag))
+        tag_keys = list(map(lambda d: d.get("key"), tags))
+        if len(set(tag_keys)) != len(tag_keys):
```
hackaugusto, line 2055: This can be done with a [Counter](https://docs.python.org/3/library/collections.html#collections.Counter):

```python
repeated_keys = [key for key, count in Counter(tag_keys).items() if count > 1]
```

aiven/client/cli.py
```diff
@@ -2020,6 +2049,14 @@ def service__task_create(self):
     )
     def service__topic_create(self):
         """Create a Kafka topic"""
+
+        tags = list(map(parse_tag_str, self.args.topic_option_tag))
```
hackaugusto, line 2053: nitpick: I think list comprehensions are a bit more idiomatic:

```python
all_tags = [parse_tag_str(tag_arg) for tag_arg in self.args.topic_option_tag]
tag_keys = [tag["key"] for tag in all_tags] 
```

aiven/client/cli.py
```diff
@@ -2040,9 +2078,28 @@ def service__topic_create(self):
...
+            for k in set(tag_keys):
+                tag_keys.remove(k)
+            raise argx.UserError(f"Duplicate tags detected: {', '.join(tag_keys)}")
+
+        tags = None
+        if len(new_tags) + len(untags) > 0:
+            topic = self.client.get_service_topic(
+                project=self.get_project(), service=self.args.service_name, topic=self.args.topic
+            )
+            tags = topic.get("tags") or []
```
hackaugusto, line 2100: nitpick: `dict.get` has an optional second argument for the default value, this can be `topic.get("tags", [])`

aiven/client/cli.py
```diff
@@ -2040,9 +2078,28 @@ def service__topic_create(self):
...
+        untags = list(map(parse_untag_str, self.args.topic_option_untag))
+        tag_keys = list(map(lambda d: d.get("key"), new_tags)) + untags
+        if len(set(tag_keys)) != len(tag_keys):
+            for k in set(tag_keys):
+                tag_keys.remove(k)
+            raise argx.UserError(f"Duplicate tags detected: {', '.join(tag_keys)}")
+
+        tags = None
+        if len(new_tags) + len(untags) > 0:
+            topic = self.client.get_service_topic(
```
hackaugusto, line 2097: This needs a comment. Is it necessary to fetch the existing tags because `update_service_topic` overwrite all the tags instead of adding on top?

aiven/client/cliarg.py
```diff
@@ -92,6 +92,30 @@ def wrapped(self):
 arg.replication = arg("--replication", type=int, required=True, help="Replication factor")
 arg.retention = arg("--retention", type=int, help="Retention period in hours (default: unlimited)")
 arg.retention_bytes = arg("--retention-bytes", type=int, help="Retention limit in bytes (default: unlimited)")
+arg.tag = arg(
+    "--tag",
+    dest="topic_option_tag",
+    metavar="KEY[=VALUE]",
+    action="append",
+    default=[],
```
hackaugusto, line 100: Having mutable objects in globals may cause weird behavior. I think it is better to leave this with `None` as a default and handle that after parsing the CLI arguments.

aiven/client/client.py
```diff
@@ -602,7 +605,9 @@ def update_service_topic(
                 "replication": replication,
                 "retention_bytes": retention_bytes,
                 "retention_hours": retention_hours,
-            },
+            } | ({
```
hackaugusto, line 608: This doesn't seem right, dictionaries don't support he or operator:

```python
>>> d = {1:1}
>>> d2 = {2:2}
>>> d | d2
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'dict' and 'dict'
```
You probably want to use `dict.update`.

hackaugusto (changes requested):

ivanyu (dismissed):

aiven/client/client.py
tvainika: I didn't notice that is Python 3.9 new feature...

hackaugusto: And I didn't know that 3.9 introduced it :smile:  https://www.python.org/dev/peps/pep-0584/

But 3.6, 3.7, and 3.8 are supported versions:

https://github.com/aiven/aiven-client/blob/0ec1a2b02c0863d2fbfc42298d437389b1e87aac/setup.py#L50-L52

So we shouldn't use it here

aiven/client/cli.py
```diff
@@ -67,6 +68,31 @@ def convert_str_to_value(schema, str_value):
         raise argx.UserError("Support for option value type(s) {!r} not implemented".format(schema["type"]))
 
 
+tag_key_re = re.compile(r"[\w\-]+")
+tag_value_re = re.compile(r"[\w\-,. ]*")
+
+
+def parse_tag_str(kv):
+    k, v = (kv.split(sep='=', maxsplit=1) + [''])[:2]
```
hackaugusto, line 76: Does the API treat the empty string `""` and `None` the same way? If not I suppose this should be `[None]` instead of `['']` since the user can define the empty string with `tag=`

tvainika: API expects always value, but it can be an empty string. `None` (omitting `"value"`) is not allowed. So both `tag=` and `tag` will produce empty string as tag value intentionally.

hackaugusto (approved):
LGTM

PR #189, opened 2021-03-26
Add default tags for service_topic_create client call

tvainika:
Add default value for call parameters to keep client API backward compatible so this can be used as Python lib also, not only CLI tool.

### DIFF
--- a/aiven/client/client.py
+++ b/aiven/client/client.py
@@ -567,21 +567,23 @@ class AivenClient(AivenClientBase):
         retention_bytes,
         retention_hours,
         cleanup_policy,
-        tags,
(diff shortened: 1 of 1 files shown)


aiven/client/client.py
```diff
@@ -567,7 +567,7 @@ def create_service_topic(
         retention_bytes,
         retention_hours,
         cleanup_policy,
-        tags,
+        tags=[],
```
rikonen, line 570: Unsafe default argument

rikonen (changes requested):

rikonen (approved):

PR #193, opened 2021-04-08
Add support for setting group of m3 service user

fingon:


aiven/client/cli.py
```diff
@@ -1197,25 +1197,30 @@ def service__migration_status(self):
...
+    def _parse_access_control(self):
         arg_vars = vars(self.args)
-        return {
+        result = {
             key: arg_vars[key].split()
             for key in {"redis_acl_keys", "redis_acl_commands", "redis_acl_categories"}
             if arg_vars[key] is not None
         }
+        for key in ["m3_group"]:
+            if key in arg_vars:
```
heikju, line 1208: Isn't `m3_group` always in the args (with value `None` if `--m3-group` not given)? Is the idea to always include it in the result (in which case this `if` is redundant), or only when the flag is given (in which case it could be handled in the dict comprehension above)?

aiven/client/cli.py
```diff
@@ -1197,25 +1197,30 @@ def service__migration_status(self):
...
-    def _parse_redis_access_control(self):
+    def _parse_access_control(self):
         arg_vars = vars(self.args)
-        return {
+        result = {
             key: arg_vars[key].split()
             for key in {"redis_acl_keys", "redis_acl_commands", "redis_acl_categories"}
             if arg_vars[key] is not None
         }
+        for key in ["m3_group"]:
```
heikju, line 1207: Our style guide would say `in {"m3_group"}:`, though obviously here it doesn't really make a difference.

heikju (changes requested):

aiven/client/cli.py
fingon: It should probably only apply to `if key in` and not `for key in`, as the former leads to reasonable speed and the latter leads to unexpected ordering of the nested block in for.

alexole (approved):

ansrivas (approved):

PR #197, opened 2021-04-19
cli: accept token in avn login command

chornsby:
For users with platform authentication disabled the `avn login` command
cannot authenticate the aiven-client. This authentication method may
have been excluded by admins at the account level.

To authenticate the client the user could edit the JSON configuration
file to add an authentication token directly, but by accepting it also
in the `avn login` command we can make the functionality more easily
discoverable.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,6 +68,12 @@ password you use on Aiven::
 
 The command will prompt you for your password.
 
+You can also use an access token generated in the Aiven Console::
+
+  $ avn user login <you@example.com> --token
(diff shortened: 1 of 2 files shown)


rikonen (dismissed):
I think adding some `@arg("--token", action="store_true", help="Provide an access token instead of password")` parameter would be better. Guessing what the value is based on length and showing "or token" part in password prompt could be quite confusing.

rikonen (approved):

PR #199, opened 2021-04-26
Add custom datadog tags

jhonkola:
Added "--user-config-json" parameter to support json in specifying
more complex user config for commands. The json document is checked
against schema in Aiven API, cli does only basic syntactic check.

The new parameter can be added with a decorator `user_config_json()`
found in aiven/client/cliarg.py:43

Currently the parameter is implemented for service integration related
commands.

Unit tests for the decorator are in tests/test_cliarg.py

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -1530,11 +1530,14 @@ ssl.truststore.type=JKS
     @arg.project
     @arg("-d", "--endpoint-name", help="Integration endpoint name", required=True)
     @arg("-t", "--endpoint-type", help="Integration endpoint type", required=True)
+    @arg.user_config_json()
     @arg.user_config
     @arg.json
     def service__integration_endpoint_create(self):
         """Create a service integration endpoint"""
-        if self.args.user_config:
+        if self.args.user_config_json:
+            user_config = self.args.user_config_json
+        elif self.args.user_config:
             project = self.get_project()
             user_config_schema = self._get_endpoint_user_config_schema(
                 project=project, endpoint_type_name=self.args.endpoint_type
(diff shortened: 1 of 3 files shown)


aiven/client/cli.py
```diff
@@ -62,6 +62,7 @@ def convert_str_to_value(schema, str_value):
             ) from ex
     elif "array" in schema["type"]:
         return [convert_str_to_value(schema["items"], val) for val in str_value.split(",")]
+
```
rikonen, line 65: Spurious linefeed

aiven/client/cliarg.py
```diff
@@ -40,6 +40,37 @@ def wrapped(self):
     return wrapper
 
 
+def user_config_json(param_name):
```
rikonen, line 43: param_name always gets the same value. I'd just remove this parameter. This function already makes assumptions about other parameter names so might as well hard code this one too.

rikonen (changes requested):
Commits should be squashed. No point in merging the wip stuff into master.

rikonen:
Also could update pull request title not to include username

aiven/client/cliarg.py
jhonkola: `param_name` removed, parsed json is always put into destination `user_config_json`

jhonkola:
Implemented requested changes

rikonen (approved):

rikonen:
Title of the pull request was adding support for datadog tags but there wasn't anything datadog tag specific here. As a follow up it would be good to update the docs to mention how you can utilize this feature to actually set those tags.

jhonkola:
Should that go to README? I do not see any other documentation, and PyPi documentation seems to be generated from README. There is no section in README dealing with service integrations, so one should be created as well to cover the commands that the new parameter relates to at the moment.

lornajane:
Adding examples to the README would get a :+1:  from me

PR #204, opened 2021-06-11
Add support for setting Redis ACL channels

deztructor:
Provide support to provide channels' ACL for Redis users.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -1201,7 +1201,7 @@ class AivenCLI(argx.CommandLineTool):
         arg_vars = vars(self.args)
         return {
             key: arg_vars[key].split()
-            for key in {"redis_acl_keys", "redis_acl_commands", "redis_acl_categories"}
+            for key in {"redis_acl_keys", "redis_acl_commands", "redis_acl_categories", "redis_acl_channels"}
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -1430,6 +1433,7 @@ def service__user_password_reset(self):
     @arg("--redis-acl-keys", help="ACL rules for keys")
     @arg("--redis-acl-commands", help="ACL rules for commands")
     @arg("--redis-acl-categories", help="ACL rules for command categories")
+    @arg("--redis-acl-channels", help="ACL rules for channels (Redis only)")
```
ivan-savciuc, line 1436: Here we have `(Redis only)` in the end, but not arg's from above, and I wonder whether we should make it consistent with the others (adding this to every Redis related argument)?

ivan-savciuc (commented):
Otherwise LGTM

aiven/client/cli.py
deztructor: I made it consistent with other params (typo due to copy/paste from another place). Command itself is marked as Redis-only help="Set Redis service user access control").

ivan-savciuc (approved):

tsharju (approved):

PR #205, opened 2021-06-14
jhonkola: add parameter to project creation

jhonkola:
Added parameter `--use-source-project-billing-group` to control
billing group creation when doing a copy from existing project.

When used, the new project will use the same billing group as
the source project.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -3328,6 +3328,15 @@ ssl.truststore.type=JKS
         metavar="PROJECT",
         help="Copy project settings from an existing project",
     )
+    @arg(
+        "--use-source-project-billing-group",
+        action="store_true",
+        default=False,
+        help=(
+            "If copying from existing project, use the same billing group "
+            "used by source project instead of creating a new one"
+        )
+    )
     @arg.country_code
     @arg.billing_address
     @arg.billing_extra_text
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -3352,6 +3361,7 @@ def project__create(self):
                 vat_id=self.args.vat_id,
                 billing_emails=self.args.billing_email,
                 tech_emails=self.args.tech_email,
+                use_source_project_billing_group=self.args.use_source_project_billing_group
```
deztructor, line 3364: tiny nitpick: I'd put terminating comma to make potential next parameter addition one-liner.

jhonkola: Fixed

aiven/client/client.py
```diff
@@ -1397,6 +1398,8 @@ def create_project(
             body["billing_emails"] = [{"email": email} for email in billing_emails]
         if tech_emails is not None:
             body["tech_emails"] = [{"email": email} for email in tech_emails]
+        if use_source_project_billing_group is not None:
+            body["use_source_project_billing_group"] = True
```
axeoman, line 1402: Doesn't change anything in that case but it should be `body["use_source_project_billing_group"] = use_source_project_billing_group`

axeoman (changes requested):
Looks good except small inconsistency in `self.client.create_project` (see comment).

aiven/client/client.py
jhonkola: Fixed

axeoman (approved):

axeoman:
Oh, did not noticed that, commit title perhaps should be `cli: add parameter to project creation`

axeoman (approved):

PR #206, opened 2021-06-22
argx: skip all descriptors when scanning plugin objects (SRES-1381)

jriddy:

Plugins rely on the ability to delay evaluation of `@property` values until after scanning and set-up of plugin classes is complete. However, some related descriptor types like `@cached_property` are not actually instances of the `property` class at all. This has caused problems downstream with a [bug](https://aiven.atlassian.net/browse/SRES-1314) where access to potentially not configured databases fails during the execution of unrelated commands, because the scanning of the plugin classes coerces these alternative descriptors to values, forcing early evaluation of their methods.

This change generalizes the delay-of-evaluation check from an `isinstance()` to duck-typing the attribute for being a [data descriptor](https://docs.python.org/3/howto/descriptor.html#descriptor-protocol). Both `property` and `functools.cached_property` are data descriptors, so this should work for them as well as other uses of the descriptor protocol.

### DIFF
--- a/aiven/client/argx.py
+++ b/aiven/client/argx.py
@@ -8,6 +8,7 @@ import aiven.client.client
 import argparse
 import csv as csvlib
 import errno
+import functools
 import json as jsonlib
 import logging
 import os
@@ -26,6 +27,11 @@ try:
     from .version import __version__  # pylint: disable=no-name-in-module
 except ImportError:
     __version__ = "UNKNOWN"
+
+# cached_property only exists since python 3.8
+SKIP_EVALUATION_TYPES = (property, )
+if hasattr(functools, "cached_property"):
+    SKIP_EVALUATION_TYPES += (functools.cached_property, )
 
 ARG_LIST_PROP = "_arg_list"
 LOG_FORMAT = "%(levelname)s\t%(message)s"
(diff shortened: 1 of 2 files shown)


jriddy:
I was wrong about the descriptors thing. It turns out that `cached_property` is (by design) a _non-data_ descriptor, which means it only defines `__get__()`. This is fine, except that this is _also true of normal python methods_. :weary: 

I just replaced the check with one that skips evaluating `cached_property` as well if it's defined and added a test for these cases.

aiven/client/argx.py
```diff
@@ -27,6 +28,11 @@
 except ImportError:
     __version__ = "UNKNOWN"
 
+# cached_property only exists since python 3.8
+SKIP_EVALUATION_TYPES = (property, )
+if hasattr(functools, "cached_property"):
+    SKIP_EVALUATION_TYPES += (functools.cached_property, )
```
vladan, line 34: Mutating tuples feels a bit dirty, but python has no problem with it so i guess fine.
A list here would be preferable, but that's a minor thing and shouldn't hold this off.

Or, you could:
```
if hasattr(functools, "cached_property"):
    SKIP_EVALUATION_TYPES = (property, functools.cached_property)
else:
    SKIP_EVALUATION_TYPES = (property,)
```

All in all, really unimportant :)

vladan (approved):
Looks perfectly fine, but I'd rather another person through an eye at it and merge if everything's OK.

jriddy:
It's not mutating tuples, it's just a shortcut for reassigning the variable
to another tuple. Same as with arithmetic on number types

rikonen (approved):

PR #209, opened 2021-07-17
argx: add args argument to main function of argx

rominf:

This is for allowing running argx based CLI from Python console.


axeoman (approved):
LGTM

PR #210, opened 2021-07-31
argx: allow passing a parser to CommandLineTool

facetoe:
In some cases it can be useful to pass an already instantiated parser
instance to the CommandLineTool, so allow this if needed.

This can be used to extend the tool with additional commands sourced elsewhere.

### DIFF
--- a/aiven/client/argx.py
+++ b/aiven/client/argx.py
@@ -122,12 +122,12 @@ class Config(dict):
 
 
 class CommandLineTool:  # pylint: disable=old-style-class
-    def __init__(self, name):
+    def __init__(self, name, parser: argparse.ArgumentParser = None):
         self.log = logging.getLogger(name)
         self.config = None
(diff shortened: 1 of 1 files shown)


vladan (dismissed):
if we don't need typing info then lgtm

facetoe:
Added the types  👍

jsou (approved):

aiven/client/argx.py
```diff
@@ -122,12 +122,12 @@ def save(self):
 
 
 class CommandLineTool:  # pylint: disable=old-style-class
-    def __init__(self, name):
+    def __init__(self, name, parser: argparse.ArgumentParser = None):
```
vladan, line 125: This should have been Optional if the default is None.

facetoe: Ugh good point. I'll submit another PR.

PR #211, opened 2021-08-05
argx: make parser optional

facetoe:
As the default is None, this paramater should have been defined as
Optional, so fix this up.

The type hints were incorrect so fix them up.


rikonen (changes requested):
Fails static check

rikonen (approved):

PR #212, opened 2021-08-12
Add migration check support and update PG versions for upgrade_check [BF-597]

alexole:
* Added migration_check support to `service task-create` command
* Added command to check the status of a task - `service task-get`
* Updated list of supported PG versions for upgrade check (renamed target version parameter to match the style)

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -2084,9 +2084,9 @@ ssl.truststore.type=JKS
         default="upgrade_check",
     )
     @arg(
-        "--target_version",
+        "--target-version",
         help="Upgrade target version",
-        choices=["9.5", "9.6", "10", "11"],
+        choices=["10", "11", "12", "13"],
     )
     @arg("--format", help="Format string for output, e.g. '{name} {retention_hours}'")
     @arg.json
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -8,6 +8,7 @@ from aiven.client.cliarg import arg
 from aiven.client.speller import suggest
 from collections import Counter
 from decimal import Decimal
+from typing import Callable
 from urllib.parse import urlparse
 
 import errno
@@ -103,9 +104,20 @@ def optional_auth(fun):
     return fun
 
 
+if (sys.version_info.major, sys.version_info.minor) >= (3, 8):
+    from typing import Optional, Protocol
+
+    class ClientFactory(Protocol):  # pylint: disable=too-few-public-methods
+        def __call__(self, base_url: str, show_http: bool, request_timeout: Optional[int]):
+            ...
+else:
+    ClientFactory = Callable[..., client.AivenClient]
+
+
 class AivenCLI(argx.CommandLineTool):
-    def __init__(self):
+    def __init__(self, client_factory: ClientFactory = client.AivenClient):
         argx.CommandLineTool.__init__(self, "avn")
+        self.client_factory = client_factory
         self.client = None
         for plugin in PLUGINS:
             plugincli = plugin.ClientPlugin()
@@ -3828,7 +3840,7 @@ server_encryption_options:
             raise
 
     def pre_run(self, func):
-        self.client = client.AivenClient(
+        self.client = self.client_factory(
             base_url=self.args.url,
             show_http=self.args.show_http,
             request_timeout=self.args.request_timeout,
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -2080,31 +2080,84 @@ ssl.truststore.type=JKS
     @arg(
         "--operation",
         help="Task operation",
-        choices=["upgrade_check"],
+        choices=["migration_check", "upgrade_check"],
         default="upgrade_check",
     )
     @arg(
         "--target-version",
         help="Upgrade target version",
         choices=["10", "11", "12", "13"],
+        required=False,
+    )
+    @arg(
+        "--source-service-uri",
+        help="Migration: source URI for migration",
+        required=False,
+    )
+    @arg(
+        "--ignore-dbs",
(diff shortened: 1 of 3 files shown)


kmichel-aiven (changes requested):
Code looks ok but tests are missing.

aiven/client/cli.py
```diff
@@ -2080,32 +2080,85 @@ def service__topic_get(self):
...
-        choices=["upgrade_check"],
+        choices=["migration_check", "upgrade_check"],
         default="upgrade_check",
     )
     @arg(
-        "--target_version",
+        "--target-version",
         help="Upgrade target version",
-        choices=["9.5", "9.6", "10", "11"],
+        choices=["10", "11", "12", "13"],
```
janw, line 2089: I'd prefer using `aiven.shared.magic.PG_VERSIONS` here to avoid a stale in-place list in the future

```suggestion
        choices=PG_VERSIONS[1:],
```

janw (changes requested):
~Single suggestion~, LGTM.

edit: Didn't realize what repo this was in :facepalm:

aiven/client/cli.py
jankatins, line 2089: Seems no one is upgrading via the cli :-)

tests/test_cli.py
```diff
@@ -107,3 +108,78 @@ def test_create_user_config():
...
+class MockResponse:  # pylint: disable=too-few-public-methods
+    def __init__(self, data, status_code):
+        self.data = data
+        self.status_code = status_code
+
+    def json(self):
+        return self.data
+
+
+def create_patched_cli():
```
jankatins, line 122: Maybe "authenticated_cli"?

alexole: good point, spend so much on this mocking issue, so haven't even thought about that

alexole: done @jankatins

tests/test_cli.py
```diff
@@ -107,3 +108,79 @@ def test_create_user_config():
...
     }
+
+
+class MockResponse:  # pylint: disable=too-few-public-methods
+    def __init__(self, data, status_code):
+        self.data = data
+        self.status_code = status_code
+
+    def json(self):
+        return self.data
```
jankatins, line 119: Is this class used at all? The below uses only `.json()`, so basically passes back the original dict.

tests/test_cli.py
```diff
@@ -107,3 +108,79 @@ def test_create_user_config():
...
+            "create_time": "2021-08-13T08:00:45Z",
+            "result": "",
+            "success": None,
+            "task_id": "79803598-d09a-44bf-ae2b-34aad942f4e8",
+            "task_type": "mysql_migration_check"
+        }
+    }
+
+    with mock.patch("aiven.client.client.AivenClient") as cli_client_mock:
+        cli_client_mock.return_value.create_service_task.return_value = MockResponse(task_create_response, 200).json()
```
jankatins, line 145: This is equivalent to `... = task_create_response`, isn't it?

(same for below)

alexole: I should have polished it :) Yeah, not mocking class not needed. Thanks!

tests/test_cli.py
alexole: Yeah, old code. :+1:

tests/test_cli.py
```diff
@@ -107,3 +108,69 @@ def test_create_user_config():
...
+        "task": {
+            "create_time": "2021-08-13T08:00:45Z",
+            "result": "",
+            "success": None,
+            "task_id": "79803598-d09a-44bf-ae2b-34aad942f4e8",
+            "task_type": "mysql_migration_check"
+        }
+    }
+
+    with mock.patch("aiven.client.client.AivenClient") as cli_client_mock:
```
kmichel-aiven, line 135: Instead of using `mock.patch` and import paths, it would be less brittle to edit that part:
https://github.com/aiven/aiven-client/blob/1760ebfc970c00aa634db151d779165c5469f35e/aiven/client/cli.py#L106-L109
to add a `client_class = AivenClient` parameter to the init, and use it later to create the client.
this way a custom client can be injected when creating the `AivenCLI` instance in the `authenticated_cli` fixture.

jankatins (dismissed):
LGTM

tests/test_cli.py
jankatins: So basically the fixture would set `cli.client_class = Mock()` and then change the tests:

```diff
- cli_client_mock.return_value.get_service_task.return_value = task_get_response
+authenticated_cle.client_class.return_value.get_service_task.return_value = task_get_response
```
?

Looks a bit ugly, but returning a tuple `(authenticated_cli, client_mock)` looks just as ugly :-(

alexole: Frankly speaking, I don't see why mocking of client is not appropriate here. I doubt adding `AivenClient` as a parameter is justifiable here, also how will it make tests less brittle? Can you give an example please? @kmichel-aiven

kmichel-aiven: Example here : https://github.com/aiven/aiven-client/commit/7eb65a07d28111f305ac61d90bf4a3e9b3db91fb

All variants I've tried have some degrees of ugliness, most of it caused by `argx.CommandLineTool.run` doing a lot of stuff all at once, I still think it's a first step toward making AivenCLI easier to test, and tested.

alexole: Alright, I wouldn't say it look beautiful now, I'd prefer patching instead, but at this point I really don't mind. Will change as you suggest.

jankatins (approved):
LGMT, will leave the final review/merge to @kmichel-aiven

kmichel-aiven (approved):

PR #215, opened 2021-08-23
cliarg: fix false parameter conflict bug

jhonkola:
Fix bug in `--user-config-json` parameter validation that caused an error
when `-c` parameter was used even if `--user-config-json` was not.

### DIFF
--- a/aiven/client/cliarg.py
+++ b/aiven/client/cliarg.py
@@ -11,6 +11,9 @@ import os
 
 
 def get_json_config(path_or_string):
+    # If parameter is empty, return an empty dict
+    if not path_or_string:
+        return {}
     if path_or_string.startswith("@"):
         filepath = path_or_string[1:]
         with open(filepath, "r") as config_file:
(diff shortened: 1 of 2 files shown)


aiven/client/cliarg.py
```diff
@@ -46,7 +49,7 @@ def user_config_json():
     def wrapper(fun):
         arg(
             "--user-config-json",
-            default="{}",
+            default={},
```
tsharju, line 52: Imho something line `None` would make a better default here. Now the function takes in basically `dict` or `str` whereas `Optional[str]` would maybe be more consistent. I know there's no typing here but still.

tsharju (changes requested):
One comment.

aiven/client/cliarg.py
jhonkola: Fixed.

tsharju (approved):
LGTM

tvainika:
@jhonkola @tsharju Are you planning to release 2.13.2 for this? IMO it would make sense as this regression can be quite confusing.

PR #218, opened 2021-09-03
cli: add support for dynamic disk sizing

ilpon:
Add support for a new feature that allows increasing the disk space provided by the plan within the plan defined limits.

### DIFF
--- a/aiven/client/cliarg.py
+++ b/aiven/client/cliarg.py
@@ -102,6 +102,9 @@ arg.config_file = arg(
     help="Additional configuration option whose value is loaded from file in the form name=filename",
 )
 arg.country_code = arg("--country-code", help="Billing country code")
+arg.disk_space_mb = arg(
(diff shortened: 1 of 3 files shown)


aiven/client/cliarg.py
```diff
@@ -102,6 +102,9 @@ def wrapped(self):
     help="Additional configuration option whose value is loaded from file in the form name=filename",
 )
 arg.country_code = arg("--country-code", help="Billing country code")
+arg.disk_space_mb = arg(
+    "--disk-space-gib", dest="disk_space_mb", type=lambda value: int(value) * 1024, help="Disk space for data storage (GiB)"
```
sjamgade, line 106: do we want to have some kind of sanity check here ? (eg: -ve value, super large value...)
or does the API do that ?

sjamgade (approved):
LGTM
do we need to worry about rounding errors while printing also

aiven/client/cliarg.py
ilpon: Yes, API checks that the value is between the limits set by the plan.

PR #220, opened 2021-09-07
Connection information commands for kafkacat and psql

mwfrojdman:
Examples:
```
# avn service connection-info pg uri mypg  --route privatelink --replica
postgres://avnadmin:lfkjJfjQ6vQIG7Xr@privatelink-mypg.aiven.domain:21089/defaultdb?sslmode=require
# avn service connection-info kafkacat -a sasl --username mwf mykafka
kafkacat -b mykafka.aiven.domain:21100 -X security.protocol=SASL_SSL -X ssl.ca.location=ca.pem -X sasl.mechanisms=SCRAM-SHA-256 -X sasl.username=mwf -X sasl.password=eTKOcEM8ulGG4nCa
```

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -5,6 +5,9 @@
 from . import argx, client
 from aiven.client import envdefault
 from aiven.client.cliarg import arg
+from aiven.client.connection_info.common import Store
+from aiven.client.connection_info.kafka import KafkaCertificateConnectionInfo, KafkaSASLConnectionInfo
+from aiven.client.connection_info.pg import PGConnectionInfo
 from aiven.client.speller import suggest
 from collections import Counter
 from decimal import Decimal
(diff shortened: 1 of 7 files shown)


lornajane:
This is a real quality of life improvement and I think `connection-info` is good naming! We should call `kafkacat` by its new name `kcat` though, and also I have questions:
 - Is the first positional argument service type, and could we figure that out from the named service? Or is it a destination application?
 - And then if the user doesn't choose `uri`, what are the other options? Is there a default or would that even make sense since it's not-the-last of the positional arguments.
 - Do these commands also do the work of `user-creds-download` I am not sure if that would be surprising if the user was only expecting to get a string? Or maybe I should ask, which parts of the README docs should we update with this new feature https://github.com/aiven/aiven-client#id18
 - Why are both `--replica` and `--usage` supported when they do the same thing and you can't do both?

I would also like to update the usage strings but maybe when I've stopped asking so many questions.

mwfrojdman:
> This is a real quality of life improvement and I think `connection-info` is good naming! We should call `kafkacat` by its new name `kcat` though, and also I have questions:

Yeah I was thinking about this. (Old) Linux distributions will still have the command as `kafkacat`, while new installations probably use `kcat`. Maybe should add both as aliases to each other

>     * Is the first positional argument service type, and could we figure that out from the named service? Or is it a destination application?

I was thinking about that. The issue is that the arguments differ between service types. For example `--usage`/`--replica` is applicable on pg but not on kafka, and vice versa `--authentication-method` only makes sense in the context of Kafka.

Doing the argument parsing depending on service type is possible, but not without quite big changes to the `argx` helper on top of `argparse`.

>     * And then if the user doesn't choose `uri`, what are the other options? Is there a default or would that even make sense since it's not-the-last of the positional arguments.

The `service_uri` shown for all services is a proper URI for some services like pg and kafka connect, while eg. kafka and cassandra have just `hostname:port`. I think we could add a default once we have more connection info formats covered and see if it makes sense.

For pg there's `uri` (the URL), `psql` (psql command string), `string` (DSN style PG connection string). For kafka I just added `kafkacat`, but we could add Java-style configuration file support, python client arguments, and any other clients/commands that exist out there. I was also thinking of adding a `params` format for all service types that give a `connection_info_params` style key-value mapping (that can be dumped also in JSON). Those are also exposed in the console's connection info box currently.

>     * Do these commands also do the work of `user-creds-download` I am not sure if that would be surprising if the user was only expecting to get a string? Or maybe I should ask, which parts of the README docs should we update with this new feature https://github.com/aiven/aiven-client#id18
> 
>     * Why are both `--replica` and `--usage` supported when they do the same thing and you can't do both?

The idea there was that if you have assigned the desired usage to a variable you can just pass it to the command with `--usage`, but `--replica` is easier if you're just typing along instead of `--usage replica`

> I would also like to update the usage strings but maybe when I've stopped asking so many questions.

mwfrojdman:
Changed the key/cert file saving so that by default nothing is saved, with `--write` the files are saved if they don't already exist and with `--overwrite` even if they exist (last one might be useful if you're using eg. the same project or user name after deleting the previous and the credentials are no longer the same)

aiven/client/connection_info/pg.py
```diff
@@ -0,0 +1,55 @@
...
+    def from_service(cls, service, *, route, usage, username, dbname, sslmode):
+        if service["service_type"] != "pg":
+            raise ConnectionInfoError("Cannot format pg connection info for service type {service_type}".format_map(service))
+
+        info = find_component(service["components"], route=route, usage=usage)
+        host = info["host"]
+        port = info["port"]
+        user = find_user(service, username)
+        try:
+            password = user["password"]
```
rikonen, line 25: Should also check that password is not None

aiven/client/cli.py
```diff
@@ -1235,6 +1238,130 @@ def service__get(self):
...
+    @arg.project
+    @arg.service_name
+    @arg("-r", "--route", choices=("dynamic", "privatelink", "public"), default="dynamic")
+    @arg("-a", "--kafka-authentication-method", choices=("certificate", "sasl"), default="certificate")
+    @arg("-u", "--username", default="avnadmin")
+    @arg("--ca", default="ca.pem", dest="ca_path")
+    @arg("--client-cert", default="service.crt", dest="client_cert_path")
+    @arg("--client-key", default="service.key", dest="client_key_path")
+    @arg("-w", "--write", action="store_true", help="Save certificate and key files if they don't not exist")
+    @arg("-ww", "--overwrite", action="store_true", help="Save and overwrite certificate and key files")
```
rikonen, line 1262: Maybe `W` instead? Multi letter short versions can get quite confusing when you pass multiple parameters with single dash.

aiven/client/cli.py
```diff
@@ -1235,6 +1238,130 @@ def service__get(self):
...
+        print(" ".join(cmd))
+
+    @arg.project
+    @arg.service_name
+    @arg("-r", "--route", choices=("dynamic", "privatelink", "public"), default="dynamic")
+    @arg("--usage", choices=("primary", "replica"))
+    @arg("--replica", action="store_true")
+    @arg("-u", "--username", default="avnadmin")
+    @arg("-d", "--dbname", default="defaultdb")
+    @arg("--sslmode", default="require")
```
rikonen, line 1328: Having `choices` defined for this would be more user-friendly

rikonen (commented):
Few minor comments, in general looks reasonable to me

aiven/client/cli.py
mwfrojdman: Changed to `-W`

aiven/client/cli.py
mwfrojdman: Added `choices`

rikonen (approved):

rikonen (approved):

PR #221, opened 2021-09-10
Improve support around service EOL

ngilles-aiven:
This PR is intended to improve service EOL support in the cli's commands
Notably:

  * If a specific version is provided to `service create` or `service update` command some sanity checks are peformed, checking EOL status of the service  and asking for confirmation if needed before proceeding
  * `service list` command now shows EOL notifications

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -947,6 +947,7 @@ class AivenCLI(argx.CommandLineTool):
         "plan",
         "create_time",
         "update_time",
+        "notifications",
     ]]
     EXT_SERVICE_LAYOUT = ["service_uri", "disk_space_mb", "user_config.*", "databases", "users"]
 
@@ -967,6 +968,16 @@ class AivenCLI(argx.CommandLineTool):
         layout = self.SERVICE_LAYOUT[:]
         if self.args.verbose:
             layout.extend(self.EXT_SERVICE_LAYOUT)
+
+        # Format service notifications
+        for service in services:
+            for notification in service["service_notifications"]:
+                service.setdefault("notifications", [])
+                if notification["type"] == "service_end_of_life":
+                    # Python can't proper parse Z terminated ISO strings
+                    eol_time = parse_iso8601(notification["metadata"]["service_end_of_life_time"])
+                    eol_date = eol_time.date().isoformat()
+                    service["notifications"].append(f"EOL: {eol_date} Upgrade available")
 
         self.print_response(services, format=self.args.format, json=self.args.json, table_layout=layout)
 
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -33,6 +34,8 @@
 ]
 PLUGINS = []
 
+EOL_ADVANCED_WARNING_TIME = timedelta(weeks=26)  # Give 6 months advance notice for EOL services
```
fingon, line 37: nit: Should be `ADVANCE_` not `ADVANCED_`

aiven/client/cli.py
```diff
@@ -255,6 +275,21 @@ def enter_password(self, prompt, var="AIVEN_PASSWORD", confirm=False):
...
 
+    def print_boxed(self, lines: List[str]) -> None:
+        longest = max(len(line) for line in lines)
+
+        print("*" * longest)
+        for line in lines:
+            print(line)
+        print("*" * longest)
+
+    def confirm(self, prompt: str = "confirm? "):
```
fingon, line 286: This could have some hint also (e.g. `.. (y/N)? ` as default seems to be effectively N).

aiven/client/cli.py
```diff
@@ -3382,6 +3431,95 @@ def _get_service_type_user_config_schema(self, project, service_type):
...
+            if len(services_matching_version) == 0:
+                raise argx.UserError(
+                    f"{service_type} v{requested_version} is reaching EOL and cannot be used for new services"
+                )
+
+            # ...But should be discouraged
+            self.print_boxed(warning)
+
+            if not self.confirm("continue anyway? "):
+                raise argx.UserError("Aborted")
```
fingon, line 3515: I wouldn't put these absolute checks here, as there may be cases where e.g. customer specific dispensation is made to allow for it later on and clients are updated seldom. Instead, similar checks should exist in the backend which is updated constantly.

Due to that, any condition here should be just warning with confirm prompt.

fingon (dismissed):
Overall LGTM.

aiven/client/cli.py
ngilles-aiven: To be honest, I wasn't fully happy with these either because, as you say, these should be in sync with the "real" rules are, as enforced by server side, which may change at any point (and considering aiven-client installations in the wild could be "old" ....)
Should we still have a difference in messaging? something like:
1)  "advance warning" -> consider deploying a newer version
2) "passed availability" -> warning, operation may fail?

or reduce it just to the 1) ?

fingon: I don't have strong opinion either way - 2 adds some extra value, but is added complexity on this side (and something that has to be kept up to date). By default I'd prefer probably simpler choice (just 1), but I won't object to 2 either as long as it doesn't block operation.

ngilles-aiven: Went with option 1 ... KISS prevails

fingon (approved):
LGTM.

PR #222, opened 2021-09-20
flink: add supported methods

HelenMel:
__WHY__:
New Flink service will be released soon

__WHAT__:
Add Flink CLI support for existing REST API endpoints.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -3062,24 +3062,24 @@ ssl.truststore.type=JKS
     @arg.project
     @arg.service_name
     def service__flink__table__list(self):
-        """List created Flink tables"""
+        """List Flink tables"""
         project_name = self.get_project()
         self.print_response(self.client.list_flink_tables(project_name, self.args.service_name))
 
     @arg.project
     @arg.service_name
     @arg("integration_id", help="Service integration ID")
-    @arg("-n", "--table_name", required=True, help="Table name")
-    @arg("-kt", "--kafka_topic", required=False, help="Topic name, used as a source/sink. (Kafka integration only)")
-    @arg("-dt", "--jdbc_table", required=False, help="Table name in Database, used as a source/sink. (PG integration only)")
-    @arg("-p", "--partitioned_by", required=False, help="A column from a schema, table will be partitioned by")
+    @arg("-n", "--table-name", required=True, help="Table name")
+    @arg("--kafka-topic", required=False, help="Topic name, used as a source/sink. (Kafka integration only)")
+    @arg("--jdbc-table", required=False, help="Table name in Database, used as a source/sink. (PG integration only)")
+    @arg("-p", "--partitioned-by", required=False, help="A column from a schema, table will be partitioned by")
     @arg(
         "-l",
-        "--like_options",
+        "--like-options",
         required=False,
         help="Clause can be used to create a table based on a definition of an existing table"
     )
-    @arg("-s", "--schema_sql", required=True, help="Source/Sink table schema")
+    @arg("-s", "--schema-sql", required=True, help="Source/Sink table schema")
     def service__flink__table__create(self):
         """Create a Flink table"""
         project_name = self.get_project()
@@ -3124,7 +3124,14 @@ ssl.truststore.type=JKS
     @arg.project
     @arg.service_name
     @arg("job_name", help="Job Name")
-    @arg("-t", "--tables", required=False, help="List of tables required in job runtime")
+    @arg(
+        "-t",
+        "--tables",
+        nargs="*",
+        default=[],
+        required=False,
+        help="List of tables required in job runtime, e.g. table1 table2 table3"
+    )
     @arg("-s", "--statement", required=True, help="Job SQL statement")
     def service__flink__job__create(self):
         """Create a Flink job"""
@@ -3135,7 +3142,7 @@ ssl.truststore.type=JKS
                 self.args.service_name,
                 statement=self.args.statement,
                 job_name=self.args.job_name,
-                tables=self.args.tables.split(","),
+                tables=self.args.tables,
             )
         )
 
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -3061,10 +3061,18 @@ ssl.truststore.type=JKS
 
     @arg.project
     @arg.service_name
+    @arg.json
     def service__flink__table__list(self):
         """List Flink tables"""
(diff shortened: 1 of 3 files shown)


aiven/client/cli.py
```diff
@@ -3059,6 +3059,86 @@ def mirrormaker__replication_flow__delete(self):
...
+        project_name = self.get_project()
+        self.print_response(self.client.list_flink_tables(project_name, self.args.service_name))
+
+    @arg.project
+    @arg.service_name
+    @arg("integration_id", help="Service integration ID")
+    @arg("-n", "--table_name", required=True, help="Table name")
+    @arg("-kt", "--kafka_topic", required=False, help="Topic name, used as a source/sink. (Kafka integration only)")
+    @arg("-dt", "--jdbc_table", required=False, help="Table name in Database, used as a source/sink. (PG integration only)")
+    @arg("-p", "--partitioned_by", required=False, help="A column from a schema, table will be partitioned by")
```
tvainika, line 3075: For consistency long arguments should be hyphen separated, not underscore. There are more than these below. 

```suggestion
    @arg("-n", "--table-name", required=True, help="Table name")
    @arg("-kt", "--kafka-topic", required=False, help="Topic name, used as a source/sink. (Kafka integration only)")
    @arg("-dt", "--jdbc-table", required=False, help="Table name in Database, used as a source/sink. (PG integration only)")
    @arg("-p", "--partitioned-by", required=False, help="A column from a schema, table will be partitioned by")
```

aiven/client/cli.py
```diff
@@ -3059,6 +3059,86 @@ def mirrormaker__replication_flow__delete(self):
             self.args.target_cluster,
         )
 
+    @arg.project
+    @arg.service_name
+    def service__flink__table__list(self):
+        """List created Flink tables"""
```
tvainika, line 3065:
```suggestion
        """List Flink tables"""
```

created seems extra word here

aiven/client/cli.py
```diff
@@ -3059,6 +3059,86 @@ def mirrormaker__replication_flow__delete(self):
...
+        self.print_response(self.client.delete_flink_table(
+            project_name,
+            self.args.service_name,
+            self.args.table_id,
+        ))
+
+    @arg.project
+    @arg.service_name
+    @arg("job_name", help="Job Name")
+    @arg("-t", "--tables", required=False, help="List of tables required in job runtime")
```
tvainika, line 3127: This could have example or something to say this is comma separated list?
```suggestion
    @arg("-t", "--tables", required=False, help="List of tables required in job runtime, e.g. table1,table2,table3")
```

tvainika (changes requested):
Looks good, some minor observations for consistency and clarity

aiven/client/cli.py
```diff
@@ -3059,6 +3059,86 @@ def mirrormaker__replication_flow__delete(self):
...
+        """List created Flink tables"""
+        project_name = self.get_project()
+        self.print_response(self.client.list_flink_tables(project_name, self.args.service_name))
+
+    @arg.project
+    @arg.service_name
+    @arg("integration_id", help="Service integration ID")
+    @arg("-n", "--table_name", required=True, help="Table name")
+    @arg("-kt", "--kafka_topic", required=False, help="Topic name, used as a source/sink. (Kafka integration only)")
+    @arg("-dt", "--jdbc_table", required=False, help="Table name in Database, used as a source/sink. (PG integration only)")
```
rikonen, line 3074: In addition to the hyphen comment, it's better to avoid multi-letter shorthand versions. In this case perhaps just omit the shorthand versions altogether.

aiven/client/cli.py
rikonen: Instead of parsing manually from comma separated list using `nargs="*"` would be better

aiven/client/cli.py
HelenMel: yes. But only if it is cli argument name. I believe `integration_id` stays as it is

aiven/client/cli.py
```diff
@@ -3059,6 +3059,93 @@ def mirrormaker__replication_flow__delete(self):
             self.args.target_cluster,
         )
 
+    @arg.project
+    @arg.service_name
+    def service__flink__table__list(self):
+        """List Flink tables"""
+        project_name = self.get_project()
+        self.print_response(self.client.list_flink_tables(project_name, self.args.service_name))
```
tvainika, line 3067: Now this outputs JSON by default. Comparing to others list commands of `avn` I think this should have `@arg.json` and by default print human readable table. Same goes for flink table get

aiven/client/cli.py
```diff
@@ -3059,6 +3059,93 @@ def mirrormaker__replication_flow__delete(self):
...
+                self.args.kafka_topic,
+                self.args.jdbc_table,
+                self.args.partitioned_by,
+                self.args.like_options,
+            )
+        )
+
+    @arg.project
+    @arg.service_name
+    @arg("table_id", help="Table ID")
```
tvainika, line 3102: Just observation from user experience, it would be nice to be able to get table by table name, not by table id, but this might need changes with API?

aiven/client/cli.py
```diff
@@ -3059,6 +3059,93 @@ def mirrormaker__replication_flow__delete(self):
...
+    def service__flink__table__delete(self):
+        """Delete a Flink table"""
+        project_name = self.get_project()
+        self.print_response(self.client.delete_flink_table(
+            project_name,
+            self.args.service_name,
+            self.args.table_id,
+        ))
```
tvainika, line 3122: It would be nice to print something like `deleted` (on its own line like `service topic-delete`). 
 It seems this might also need some API changes as API for Kafka topics returns json object with `{"message": "deleted"}`

tvainika (changes requested):
There does not seem to be commands for listing jobs or deleting jobs.

It seems that after I really tested using this, got some more observations about consistency and usability. Let's take in private if you think something needs to go in sooner and others left out for later enhancements.

aiven/client/cli.py
HelenMel: Yes. It is require a proper Metadata Table, that is in scope next month

aiven/client/cli.py
HelenMel: Fair

HelenMel: I add some temporary fix, that do not need a changes from API

tvainika (approved):

PR #235, opened 2021-11-02
Fix arg type to support card CVC with leading zero

heikju:
Card security code can start with zero, so handling it as integer is not
appropriate. Use string so that leading zeroes are preserved.

Allows adding card when CVC starts with zero


tapioaiven (approved):
lgtm!

PR #236, opened 2021-11-03
Add --use-karapace to service update args

tvainika:

Kafka service can be switched to Karapace only for REST API and schema registry if not yet migrated.


ivanyu (changes requested):
Blocking for an internal discussion

ivanyu (approved):

PR #239, opened 2021-11-11
add support for unpinned connection pools [BF-552]

carobme:
We support unpinned connection pools (reusing the pool user to connect
to the database) on the platform for a while now. This adds client support
for creating unpinned pools and for unpinning existing pools.

### DIFF
--- a/aiven/client/client.py
+++ b/aiven/client/client.py
@@ -18,6 +18,8 @@ try:
 except ImportError:
     __version__ = "UNKNOWN"
 
+UNCHANGED = object()  # used as a sentinel value
+
 
 class Error(Exception):
     """Request error"""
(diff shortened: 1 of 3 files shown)


aiven/client/client.py
```diff
@@ -248,7 +250,7 @@ def update_service_connection_pool(
     ):
         body = {}
         if username is not None:
-            body["username"] = username
+            body["username"] = None if username == "" else username
```
kmichel-aiven, line 253: It could be better to keep the "empty string means set to None" in `cli.py` and not the `AivenClient`.

This would mean we can't use `None` as the default value to detect "parameter unspecified" here, and would need a sentinel object:
```python
    UNCHANGED = object()
    def update_service_connection_pool(
        self,
        project,
        service,
        pool_name,
        dbname=None,
        username=UNCHANGED,
        pool_size=None,
        pool_mode=None,
    ):
        body = {}
        if username is not UNCHANGED:
            body["username"] = username
```
Overall this will be more verbose (the cli.py part will be less trivial), but making the `"" to None` conversion on the cli side will help us write tests similar to https://github.com/aiven/aiven-client/blob/master/tests/test_cli.py#L123 . Having logic inside `AivenClient` is harder to test.

carobme: Yes, that's probably better. Changed and added some tests.

kmichel-aiven (approved):

PR #241, opened 2021-11-18
flink: support new versions of API methods [HH-690]

HelenMel:
In Flink backend were added new options for Flink table creations:
- key format for kafka
- value format for kafka
- kafka conntector type
- startup mode for kafka
- detaled columns type for tables

Colums info is in a json array and look a bit ugly and make reading harder in a table layout.
That is why, I did incude colums when customer request one table info or create a table but excluded from a table lists:

With columns:

Screenshot from 2021-11-18 09-53-43

Without columns:
Screenshot from 2021-11-18 09-53-28

Without columns it is easier to read table ids and schema.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -3120,6 +3120,7 @@ ssl.truststore.type=JKS
             "integration_id",
             "table_id",
             "table_name",
+            "schema_sql",
         ]]
         self.print_response(
             self.client.list_flink_tables(project_name, self.args.service_name), json=self.args.json, table_layout=layout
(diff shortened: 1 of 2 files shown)


juha-aiven:
The PR description should be something like "flink: support new versions of API methods". There are both additions and removals and not only for table creation but for other endpoints too.

juha-aiven:
> Colums info is in a json array and look a bit ugly and make reading harder in a table layout. That is why, I did incude colums when customer request one table info or create a table but excluded from a table lists:

It looks like the parameters are the same in the screenshots, i.e. `table list`. The other ss is old? According to the description table list doesn't have columns, but the 1st screenshot has those.

HelenMel:
@juha-aiven No, Screenshots are not old. It is stated in description, this is how it was intended

juha-aiven (approved):
I'll approve, it's a minor thing that the description and the subject could be more exact.

PR #244, opened 2021-12-17
Add command to acknowledge certificate update

jongiddy:

When a user certificate is renewed we email the project contacts to advise them to update the certificates in their Kafka clients. WE may do this periodically. There is a new API call to indicate that this step has been done and avoid follow-up emails. This PR adds a CLI command to invoke the API. The command is:
```
avn service user-creds-acknowledge ${SERVICE_NAME} --username ${USERNAME}
```

This name matches the CLI command to download the credentials: `user-creds-download`

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -1894,6 +1894,18 @@ ssl.truststore.type=JKS
     @arg.project
     @arg.service_name
     @arg("--username", help="Service user username", required=True)
+    @arg.json
+    def service__user_creds_acknowledge(self):
+        """Acknowledge that service user certificates have been updated"""
+        self.client.acknowledge_service_user_certificate(
+            project=self.get_project(),
+            service=self.args.service_name,
+            username=self.args.username,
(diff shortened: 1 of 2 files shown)


ivanyu:
The implementation LGTM, I only wonder about the user-facing command name. Should it also has the "acknowledge" verb, also to be in line with other service user commands like `service user-password-reset` and `service user-set-access-control`?

jongiddy:
Happy to change the name. I chose `updated` over `acknowledge` mainly because it is easier to type without spelling mistakes.  But `user-creds-acknowledge` is more consistent.

Not sure where the `user-password-reset` and `user-set-access-control` fit in? Do you mean that the last word is an imperative verb, a command?

ivanyu:
> Not sure where the `user-password-reset` and `user-set-access-control` fit in? Do you mean that the last word is an imperative verb, a command?

Yeah, I mean this

ivanyu (approved):

PR #249, opened 2022-01-10
pg: don't hard code major versions in upgrade_check task

carobme:
Instead of hard coding the supported versions just make sure the version
consists of digits only. The API will not accept invalid versions anyway.
This will allow performing the upgrade_check for - and subsequently
upgrading to - pg14 (and upcoming versions, as soon as they are offered).

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -2498,7 +2498,7 @@ ssl.truststore.type=JKS
     @arg(
         "--target-version",
         help="Upgrade target version",
-        choices=["10", "11", "12", "13"],
+        type=int,
         required=False,
     )
     @arg(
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -2515,7 +2515,7 @@ def service__task_create(self):
                 raise argx.UserError("--target-version is required for this operation")
             body = {
                 "task_type": self.args.operation,
-                "target_version": self.args.target_version,
+                "target_version": str(self.args.target_version),
```
laysauchoa, line 2518: so is it like it will be an integer as type is set to integer and then converted to the string to work here? what happens if it is given a wrong version?

laysauchoa (commented):
question 🙋🏻‍♀️

rominf (approved):

rominf (changes requested):
@carobme The commit/PR messages are misleading. The command that you've fixed is not about performing an upgrade; it's for checking the possibility of it. Could you please adjust the messages?

rominf (approved):

aiven/client/cli.py
laysauchoa: thanks for clarifying it!

PR #247, opened 2021-12-23
Show an error when project is not supplied

rominf:
# About this change: What it does, why it matters

Fix #246

The result of `get_project` is `None` if user didn't login, and specifies `--auth-token`. Check and raise an exception.

PR #255, opened 2022-01-17
Do not raise an error about the missing project in non-auth commands

rominf:
https://github.com/aiven/aiven-client/pull/247 introduced a bug: an error about a missing project was raised even if the project was optional.
This change fixes this by splitting get_project to get_non_none_project and get_project.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -294,12 +294,12 @@ class AivenCLI(argx.CommandLineTool):
         answer = input(prompt)
         return is_truthy(answer)
 
-    def get_project(self):
+    def get_project(self, raise_if_none=True):
         """Return project given as cmdline argument or the default project from config file"""
         if getattr(self.args, "project", None) and self.args.project:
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -294,12 +294,12 @@ def confirm(self, prompt: str = "confirm (y/N)? "):
         answer = input(prompt)
         return is_truthy(answer)
 
-    def get_project(self):
+    def get_project(self, raise_if_none=True):
         """Return project given as cmdline argument or the default project from config file"""
         if getattr(self.args, "project", None) and self.args.project:
             return self.args.project
```
dmitrii-vasilev, line 300: Can it be `None` as well?

dmitrii-vasilev: Ah, didn't check the condition properly :-)

dmitrii-vasilev (approved):
LGTM

Issue #258, opened 2022-01-17
Add type annotations


rominf:
# What is currently missing?

Type annotations (and type checks) are missing.

# How could this be improved?

Adding type annotations is beneficial for improving reliability and code maintainability. Annotations can be added gradually, module by module. Of course, these annotations should be checked using the type checker. The most popular type checker is MyPy; let's use it.

- [x] Introduce mypy (#259)
- [x] `aiven/__init__.py` (was already annotated)
- [x] `aiven/client/__init__.py` (was already annotated)
- [x] `aiven/client/__main__.py` (#259)
- [x] `aiven/client/argx.py` (#272)
- [x] `aiven/client/cli.py` (#286)
- [x] `aiven/client/cliarg.py` (#278)
- [x] `aiven/client/client.py` (#276)
- [x] `aiven/client/common.py` (#270)
- [x] `aiven/client/connection_info/` (#278)
- [x] `aiven/client/envdefault.py` (#270)
- [x] `aiven/client/pretty.py` (#281)
- [x] `aiven/client/session.py` (#270)
- [x] `aiven/client/speller.py` (#279)
- [x] `aiven/client/version.py` (#270)
- [x] `tests/__init__.py` (#283)
- [x] `tests/test_argx.py` (#283)
- [x] `tests/test_cli.py` (#283)
- [x] `tests/test_cliarg.py` (#283)
- [x] `tests/test_pretty.py` (#281)
- [x] `tests/test_session.py` (#283)
- [x] `tests/test_speller.py` (#279)
- [ ] `version.py` (#289)
- [x] Enable type checks without ignores

# Is this a feature you would work on yourself?

- [x] I plan to open a pull request for this feature

rominf:
Depends on #269.

PR #259, opened 2022-01-17
typing: add type checks for the first module

rominf:
This change is the beginning of fully annotating Aiven Client. Install
mypy, add mypy config file with everything ignored, except for the one
module, add target to Makefile, enable checks in GitHub Actions.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -38,6 +38,9 @@ jobs:
       - id: flake8
         run: make flake8
 
+      - id: mypy
+        run: make mypy
+
       - id: pyLint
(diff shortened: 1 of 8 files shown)


laysauchoa:
Super cool! 🎈  What do you think about adding some steps in some issues maybe for module that needs to be changed? Just an idea to create some smaller work packages for implementing the typing process.

rominf:
I think that putting the checklist to the issue is a good compromise between creating issues for each module and not specifying the steps at all. Will do it tomorrow.

mhoffm-aiven (approved):
lgtm

PR #264, opened 2022-01-22
Increase complexity threshold when linting

vladan:

Building packages for Fedora 35 (`python 3.10.2`) breaks with:
```
+ make test
make[3]: Entering directory '/var/home/vladan/rpmbuild/BUILD/aiven-client'
fatal: not a git repository (or any parent up to mount point /var/home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
fatal: not a git repository (or any parent up to mount point /var/home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
fatal: not a git repository (or any parent up to mount point /var/home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
python3 -m flake8 aiven tests
aiven/client/client.py:1619:5: C901 'AivenClient.create_project' is too complex (12)
aiven/client/client.py:1675:5: C901 'AivenClient.update_project' is too complex (12)
aiven/client/client.py:1854:5: C901 'AivenClient.create_billing_group' is too complex (13)
aiven/client/client.py:1908:5: C901 'AivenClient.update_billing_group' is too complex (14)
aiven/client/cli.py:205:5: C901 'AivenCLI.create_user_config' is too complex (13)
aiven/client/cli.py:846:5: C901 'AivenCLI.service__plans' is too complex (12)
aiven/client/cli.py:1840:5: C901 'AivenCLI.service__user_creds_download' is too complex (11)
aiven/client/cli.py:3743:5: C901 'AivenCLI.service__create' is too complex (11)
aiven/client/cli.py:3979:5: C901 'AivenCLI.service__update' is too complex (12)
aiven/client/argx.py:212:5: C901 'CommandLineTool.print_response' is too complex (13)
aiven/client/pretty.py:86:1: C901 'yield_table' is too complex (25)
aiven/client/connection_info/_utils.py:25:1: C901 'format_uri' is too complex (11)
make[3]: *** [Makefile:30: flake8] Error 1
make[3]: Leaving directory '/var/home/vladan/rpmbuild/BUILD/aiven-client'
error: Bad exit status from /var/tmp/rpm-tmp.YPXhnf (%check)
```

This change increases the complexity threshold and removes the linting in the worst offender (`yield_table`), hence lets the build pass successfully generating an RPM. It also adds Python 3.10 to the build matrix.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -2,3 +2,4 @@
 max-line-length = 125
 # These codes tend to disagree with yapf
 ignore = E129,W503
+max-complexity = 15
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -49,7 +49,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, pypy3]
+        python-version: ['3.7', '3.8', '3.9', '3.10', 'pypy3']
         os: [ubuntu-latest, windows-latest]
 
     steps:
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,8 @@ all: $(generated)
 aiven/client/version.py: .git/index
 	$(PYTHON) version.py $@
 
-test: flake8 mypy pylint pytest
+test: pytest
+lint: flake8 mypy pylint
 
 reformat:
 	$(PYTHON) -m isort --recursive $(PYTHON_DIRS)
--- a/aiven/client/pretty.py
+++ b/aiven/client/pretty.py
@@ -83,7 +83,7 @@ def flatten_list(complex_list):
     return flattened_list
 
 
-def yield_table(result, drop_fields=None, table_layout=None, header=True):
+def yield_table(result, drop_fields=None, table_layout=None, header=True):  # noqa
     """
     format a list of dicts in a nicer table format yielding string rows
 


mhoffm-aiven:
Why does it call a linter when building packages in the check step though? Maybe its better to have `make lint` and `make test` and seperate the linter calls into the `make lint` target. I can see value in the complaint.

rominf (changes requested):
Instead of avoiding getting linting errors, it's better to just fix these functions. They are indeed complex and fixable...

rikonen:
I don't think refactoring code should be pre-requisite for supporting new Fedora version. If we didn't enforce cyclomatic complexity before then I think correct fix is to start by ignoring the errors and if we want to start enforcing it then make another pull request that enables the check and fixes the issues.

mhoffm-aiven:
> I don't think refactoring code should be pre-requisite for supporting new Fedora version. If we didn't enforce cyclomatic complexity before then I think correct fix is to start by ignoring the errors and if we want to start enforcing it then make another pull request that enables the check and fixes the issues.

As far as i understand we did enforce it but some limit changed when using the new python version. Thats why i want to suggest dropping linting checks as a prerequisite of building the package, but keeping them as a CI check.

vladan:
Having lints removed from the `test` target really makes sense, it is a common practice to run tests while building the package, but linting shouldn't be of concern at that point.

On a side note, I think we should run all released python versions (e.g. python 3.11 as well) in the CI, regardless if it's distro packaged, or not.

rikonen:
`make rpm` shouldn't run tests at all imo but the `test` target looks somewhat reasonable, runs all types of tests including static checks.

If some default changes then I wonder is there config option we can set to explicitly define what's the current max complexity instead of having checks fail randomly.

mhoffm-aiven (approved):
lgtm

PR #270, opened 2022-02-07
Add type annotation for some files

laysauchoa:
Add type annotation on the following
files:
	aiven/client/common.py
	aiven/client/envdefault.py
	aiven/client/session.py
        aiven/client/version.py

Signed-off-by: laysa.uchoa <laysa.uchoa@aiven.io>
Co-authored-by: Roman Inflianskas <rominf@aiven.io>

It is related to this Issue: https://github.com/aiven/aiven-client/issues/258

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,15 +4,7 @@
 python_version = 3.7
 warn_redundant_casts = True
 
-# Disable errors on the code which is not annotated yet
-
 [mypy-aiven.client.*]
-ignore_errors = True
-
-[mypy-tests.*]
-ignore_errors = True
-
-[mypy-aiven.client.__main__]
 ignore_errors = False
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
@@ -23,4 +15,75 @@ warn_no_return = True
 warn_unreachable = True
 strict_equality = True
 ignore_missing_imports = True
+
+[mypy-tests.*]
+ignore_errors = False
+disallow_untyped_defs = True
+disallow_incomplete_defs = True
+check_untyped_defs = True
+no_implicit_optional = True
+warn_unused_ignores = True
+warn_no_return = True
+warn_unreachable = True
+strict_equality = True
+ignore_missing_imports = True
+
+# Disable errors on the code which is not annotated yet
+
+[mypy-aiven.client.argx]
+ignore_errors = True
+
+[mypy-aiven.client.cli]
+ignore_errors = True
+
+[mypy-aiven.client.cliarg]
+ignore_errors = True
+
+[mypy-aiven.client.client]
+ignore_errors = True
+
+[mypy-aiven.client.connection_info.__init__]
+ignore_errors = True
+
+[mypy-aiven.client.connection_info._utils]
+ignore_errors = True
+
+[mypy-aiven.client.connection_info.common]
+ignore_errors = True
+
+[mypy-aiven.client.connection_info.kafka]
+ignore_errors = True
+
+[mypy-aiven.client.connection_info.pg]
+ignore_errors = True
+
+[mypy-aiven.client.pretty]
+ignore_errors = True
+
+[mypy-aiven.client.speller]
+ignore_errors = True
+
+[mypy-tests.__init__]
+ignore_errors = True
+
+[mypy-tests.test_argx]
+ignore_errors = True
+
+[mypy-tests.test_cli]
+ignore_errors = True
+
+[mypy-tests.test_cliarg]
+ignore_errors = True
+
+[mypy-tests.test_pretty]
+ignore_errors = True
+
+[mypy-tests.test_session]
+ignore_errors = True
+
+[mypy-tests.test_speller]
+ignore_errors = True
+
+[mypy-version]
+ignore_errors = True
 
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ clean:
 
 build-dep-fedora:
 	sudo dnf install -y --best --allowerasing python3-devel python3-flake8 python3-requests \
-		tar rpmdevtools python3-mypy python3-pylint python3-isort python3-pytest
+		tar rpmdevtools python3-mypy python3-types-requests python3-pylint python3-isort python3-pytest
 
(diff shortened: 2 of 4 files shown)


aiven/client/envdefault.py
```diff
@@ -7,14 +7,21 @@
...
-AIVEN_AUTH_TOKEN = os.environ.get("AIVEN_AUTH_TOKEN")
-AIVEN_CA_CERT = os.environ.get("AIVEN_CA_CERT")
-AIVEN_CLIENT_CONFIG = os.environ.get("AIVEN_CLIENT_CONFIG", os.path.join(AIVEN_CONFIG_DIR, "aiven-client.json"))
-AIVEN_CREDENTIALS_FILE = os.environ.get("AIVEN_CREDENTIALS_FILE", os.path.join(AIVEN_CONFIG_DIR, "aiven-credentials.json"))
-AIVEN_PROJECT = os.environ.get("AIVEN_PROJECT")
-AIVEN_WEB_URL = os.environ.get("AIVEN_WEB_URL")
+AIVEN_AUTH_TOKEN: Optional[str] = os.environ.get("AIVEN_AUTH_TOKEN")
+AIVEN_CA_CERT: Optional[str] = os.environ.get("AIVEN_CA_CERT")
+AIVEN_CLIENT_CONFIG: Optional[str] = os.environ.get(
+    "AIVEN_CLIENT_CONFIG", os.path.join(AIVEN_CONFIG_DIR, "aiven-client.json")  # type: ignore
```
laysauchoa, line 21: Got some errors here and could not figure it out:

```
aiven/client/envdefault.py:24: error: Argument 1 to "join" has incompatible type "Optional[str]"; expected "Union[str, PathLike[str]]"
```
 I have tried: using `Optional[str]`, `Union[str, PathLike[str]]` and `cast()` to `str`.

related to this maybe: https://stackoverflow.com/questions/60285688/python-typehint-for-os-getenv-causes-downstream-incompatible-type-errors

if someone has some ideas what to try here, I'd highly appreciate

aiven/client/envdefault.py
```diff
@@ -7,14 +7,21 @@
 """
 
 import os
+from typing import Optional, Union
 
-USER_HOME = os.path.expanduser("~")
+USER_HOME: str = os.path.expanduser("~")
 
-AIVEN_CONFIG_DIR = os.environ.get("AIVEN_CONFIG_DIR", os.path.join(USER_HOME, ".config", "aiven"))
+AIVEN_CONFIG_DIR: Optional[str] = os.environ.get(
```
rominf, line 14: It should be just `str`, since the right part always evaluate to `str`.

aiven/client/envdefault.py
rominf: See the comment: https://github.com/aiven/aiven-client/pull/270/files#r800687065

mypy.ini
```diff
@@ -24,3 +24,51 @@ warn_unreachable = True
...
+ignore_errors = False
+disallow_untyped_defs = True
+disallow_incomplete_defs = True
+check_untyped_defs = True
+no_implicit_optional = True
+warn_unused_ignores = True
+warn_no_return = True
+warn_unreachable = True
+strict_equality = True
+ignore_missing_imports = True
```
dmitrii-vasilev, line 73: I'd do an inverse of it. Set `ignore_errors = False` for `mypy-aiven.client` and then disable it for sub-modules / sub-packages for which type checking are not enabled yet. This way it will help if somebody will add a new sub-module in a meantime. And also it will be an ultimate goal to only keep `mypy-aiven.client` block at the end.

aiven/client/common.py
```diff
@@ -1,2 +1,2 @@
 # Copyright (c) 2021 Aiven, Helsinki, Finland. https://aiven.io/
-UNDEFINED = object()
+UNDEFINED: object = object()
```
dmitrii-vasilev, line 2: Probably not really needed. Did `mypy` complain about it? I think it has pretty good type inference for cases like this.

aiven/client/envdefault.py
```diff
@@ -6,15 +6,21 @@
...
 
-AIVEN_AUTH_TOKEN = os.environ.get("AIVEN_AUTH_TOKEN")
-AIVEN_CA_CERT = os.environ.get("AIVEN_CA_CERT")
-AIVEN_CLIENT_CONFIG = os.environ.get("AIVEN_CLIENT_CONFIG", os.path.join(AIVEN_CONFIG_DIR, "aiven-client.json"))
-AIVEN_CREDENTIALS_FILE = os.environ.get("AIVEN_CREDENTIALS_FILE", os.path.join(AIVEN_CONFIG_DIR, "aiven-credentials.json"))
-AIVEN_PROJECT = os.environ.get("AIVEN_PROJECT")
-AIVEN_WEB_URL = os.environ.get("AIVEN_WEB_URL")
+AIVEN_AUTH_TOKEN: Optional[str] = os.environ.get("AIVEN_AUTH_TOKEN")
+AIVEN_CA_CERT: Optional[str] = os.environ.get("AIVEN_CA_CERT")
+AIVEN_CLIENT_CONFIG: Optional[str] = os.environ.get(
```
dmitrii-vasilev, line 19: I think it should be just `str` - it has default value which can't be `None`.

In general I think `mypy` probably can infer most of the types in this module without explicit type specifications.

version.py
```diff
@@ -26,7 +26,7 @@ def get_project_version(version_file):
         if stdout:
             git_ver = stdout.splitlines()[0].strip().decode("utf-8")
             if git_ver and ((git_ver != file_ver) or not file_ver):
-                open(version_file, "w").write("__version__ = '%s'\n" % git_ver)
+                open(version_file, "w").write("__version__: str = '%s'\n" % git_ver)
```
dmitrii-vasilev, line 29: Seems like overkill. I think `mypy` can easily recognize the type in this case.

dmitrii-vasilev (changes requested):
Looks good. Just some minor comments

dmitrii-vasilev (approved):
LGTM

PR #271, opened 2022-02-09
Add option not to revoke token on logout

### LINKED ISSUE #253: Access token passed with --token flag gets revoked after logout command
# What happened?
Configuring automation to login with a token, using following command
```
/ # avn user login machine-user@company.com --token
machine-user@company.com's Aiven access token:
INFO	Aiven credentials written to: /root/.config/aiven/aiven-credentials.json
INFO	Default project set as 'test' (change with 'avn project switch <project>')
```
After other `avn` commands are successfully executed, logging out
```
/ # avn user logout
```
Token gets revoked.
https://github.com/aiven/aiven-client/blob/e224b2aa4e89edd6aa0a777da99b5fa7800d80eb/aiven/client/cli.py#L419-L422

# What did you expect to happen?
Token passed during login with `--token` flag to not be revoked.

# What else do we need to know?
Using machine user and its token for automation, currently only terraform provider. Trying to automate creation Java keystore and truststore with `avn service user-kafka-java-creds`. I did not expect automation token to be revoked after logout. 
```
/ # avn --version
aiven-client 2.14.5
```

rominf:

Keep the current logout behavior by default, but make it possible not to
revoke the token on logout.

```
$ avn user logout --no-token-revoke
```

Fix #253.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -415,10 +415,11 @@ class AivenCLI(argx.CommandLineTool):
         else:
             self.log.info("No projects exists. You should probably create one with 'avn project create <name>'")
 
-    @arg()
+    @arg("--no-token-revoke", action="store_true", help="Do not revoke token")
     def user__logout(self):
         """Logout from current session"""
-        self.client.access_token_revoke(token_prefix=self._get_auth_token())
+        if not self.args.no_token_revoke:
+            self.client.access_token_revoke(token_prefix=self._get_auth_token())
         self._remove_auth_token_file()
 
     @arg.verbose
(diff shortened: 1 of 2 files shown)


dmitrii-vasilev (approved):
LGTM

edbighead:
Automatically revoking token after logout command is not quite intuitive and can be misinterpreted if not written in big bold red font in documentation.  

Thinking whether `avn user logout --token-revoke` is a more user friendly way ?

rominf:
@edbighead Thanks for your opinion.

> Automatically revoking token after logout command is not quite intuitive
> Thinking whether avn user logout --token-revoke is a more user friendly way ?

1. I would argue that logout that doesn't invalidate the access token doesn't sound like logout.
2. It's the current behavior, we don't want break current user setups.

> if not written in big bold red font in documentation

I agree about the point in the documentation, but I think that the red font and the size are extra. What do you think @ftisiot, @lornajane?

lornajane:
I am pretty sure our API tooling doesn't support stylesheets. The wording update would be really good to add though!

PR #272, opened 2022-02-11
Add type annotations for aiven.client.argx

rominf:

### DIFF
--- a/aiven/client/argx.py
+++ b/aiven/client/argx.py
@@ -2,8 +2,15 @@
 #
 # This file is under the Apache License, Version 2.0.
 # See the file `LICENSE` for details.
+from __future__ import annotations
+
+from .pretty import TableLayout
 from aiven.client import envdefault, pretty
-from typing import NoReturn, Optional
+from argparse import Action
+from os import PathLike
+from typing import (
+    Any, Callable, cast, Collection, Dict, List, Mapping, NoReturn, Optional, Sequence, TextIO, Tuple, Type, Union
+)
 
 import aiven.client.client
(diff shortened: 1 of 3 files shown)


rominf:
Depends on #273

aiven/client/argx.py
```diff
@@ -193,50 +200,60 @@ def add_cmds(self, add_func):
...
 
+    def _to_mapping_collection(
+        self, obj: Union[Mapping[str, Any], Collection[Mapping[str, Any]]], single_item: bool = False
+    ) -> Collection[Mapping[str, Any]]:
+        if single_item:
+            assert isinstance(obj, Mapping)
+            return [obj]
+        else:
+            assert isinstance(obj, Collection)
+            return cast(Collection[Mapping[str, Any]], obj)
```
dmitrii-vasilev, line 227: Is it possible to avoid `cast` here? The `assert` on the previous line suppose to limit the result to only one option.

dmitrii-vasilev (dismissed):
Some comments

aiven/client/argx.py
rominf: Unfortunately, mypy is not that smart:
```
aiven/client/argx.py:227: error: Incompatible return value type (got "Union[Mapping[str, Any], Collection[Mapping[str, Any]]]", expected "Collection[Mapping[str, Any]]")
```

dmitrii-vasilev: Hopefully we can fix it later

dmitrii-vasilev (approved):
LGTM

PR #273, opened 2022-02-11
Use pypy-3.7 and pypy-3.8 on CI

rominf:

`pypy3` stands for Python 3.6, which was dropped recently.


dmitrii-vasilev (approved):
LGTM

PR #274, opened 2022-02-14
drop PLUGINS global variable

mwfrojdman:
The variable doesn't seem to have any effect except polluting base
arguments in subclients. If similar functionality is needed, a
(multiple) class inheritance approach should be considered instead
(`super().method()` in all subclasses adding their plugin specific
features).


mwfrojdman:
Converted to draft until further QA is done offline

mwfrojdman:
Tested everything continues to work as intended

PR #275, opened 2022-02-18
Add support for Redis services in `cli` and `connection-info` commands

ngilles-aiven:
This PR add support for Redis services to the `cli` and `connection-info` commands to make it easier to interact with a Redis service in the command line with the help of the `avn` client.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -9,6 +9,7 @@ from aiven.client.common import UNDEFINED
 from aiven.client.connection_info.common import Store
 from aiven.client.connection_info.kafka import KafkaCertificateConnectionInfo, KafkaSASLConnectionInfo
 from aiven.client.connection_info.pg import PGConnectionInfo
+from aiven.client.connection_info.redis import RedisConnectionInfo
 from aiven.client.speller import suggest
 from collections import Counter
 from datetime import datetime, timedelta, timezone
@@ -1471,6 +1472,28 @@ class AivenCLI(argx.CommandLineTool):
         )
         print(ci.connection_string())
 
+    @arg.project
+    @arg.service_name
+    @arg("-r", "--route", choices=("dynamic", "privatelink", "public"))
+    @arg("--usage", choices=("primary", "replica"))
+    @arg("-p", "--privatelink-connection-id")
+    @arg("--replica", action="store_true")
+    @arg("-u", "--username", default="default")
+    @arg("-d", "--db")
+    def service__connection_info__redis__uri(self):
+        """Redis connection string"""
+        service = self.client.get_service(project=self.get_project(), service=self.args.service_name)
+
+        ci = RedisConnectionInfo.from_service(
+            service,
+            route=self._get_route_from_args(),
+            usage=self._get_usage_from_args(),
+            privatelink_connection_id=self._get_privatelink_connection_id_from_args(),
+            username=self.args.username,
+            db=self.args.db,
+        )
+        print(ci.uri())
+
     @optional_auth
     @arg.project
     @arg.service_name
@@ -1496,9 +1519,11 @@ class AivenCLI(argx.CommandLineTool):
             command, params, env = self._build_influx_start_info(url)
         elif service_type == "postgres":
             command, params, env = self._build_psql_start_info(url)
+        elif service_type == "rediss":
+            command, params, env = self._build_redis_start_info(url)
         else:
             raise argx.UserError(
-                "Unsupported service type {}. Only InfluxDB and PostgreSQL are supported".format(service_type)
+                "Unsupported service type {}. Only InfluxDB, PostgreSQL, and Redis are supported".format(service_type)
             )
 
         try:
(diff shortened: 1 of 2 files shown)


mypy.ini
```diff
@@ -54,6 +54,9 @@ ignore_errors = True
 [mypy-aiven.client.connection_info.pg]
 ignore_errors = True
 
+[mypy-aiven.client.connection_info.redis]
+ignore_errors = True
```
rominf, line 58: Please don't do that. We want our code base fully type annotated.

aiven/client/connection_info/redis.py
```diff
@@ -0,0 +1,53 @@
...
+from .common import ConnectionInfoError
+
+
+class RedisConnectionInfo:
+    def __init__(self, host, port, username, db, password):
+        self.host = host
+        self.port = port
+        self.username = username
+        self.password = password
+        self.db = db
```
rominf, line 11: I suggest using `@dataclass` to minimize the amount of boilerplate.

aiven/client/connection_info/redis.py
```diff
@@ -0,0 +1,53 @@
...
+            password = service['connection_info']['redis_password']
+        else:
+            user = find_user(service, username)
+            password = user.get("password")
+
+        if password is None:
+            raise ConnectionInfoError(f"Could not find password for username {username}")
+        return cls(host=host, port=port, username=username, db=db, password=password)
+
+    def params(self):
```
rominf, line 35: Consider adding `@property`.

aiven/client/connection_info/redis.py
```diff
@@ -0,0 +1,53 @@
...
+        return {
+            "host": self.host,
+            "port": self.port,
+            "user": self.username,
+            "db": self.db,
+            "password": self.password,
+        }
+
+    def uri(self):
```
rominf, line 44: Consider adding `@property`.

aiven/client/connection_info/redis.py
```diff
@@ -0,0 +1,53 @@
...
+                "Cannot format redis connection info for service type {service_type}".format_map(service)
+            )
+
+        info = find_component(
+            service["components"], route=route, usage=usage, privatelink_connection_id=privatelink_connection_id
+        )
+        host = info["host"]
+        port = info["port"]
+        if username == "default":
+            password = service['connection_info']['redis_password']
```
rominf, line 26: Please use double quotes here.

rominf (changes requested):
Not a thorough review, but some changes are needed for sure.

aiven/client/connection_info/redis.py
ngilles-aiven: Ah, I ran `yapf` but it doesn't like changing anything but whitespace... are we considering moving to `black` for the avn cli at some p oint?

aiven/client/connection_info/redis.py
ngilles-aiven: I had actually thought of something like that, but I based this code on the other "connection-info" bits, while they're mostly independent, would be nice to keep them with the same interface. Probably best to fix those as well with the same suggestions...

aiven/client/connection_info/redis.py
```diff
@@ -0,0 +1,53 @@
...
+            raise ConnectionInfoError(
+                "Cannot format redis connection info for service type {service_type}".format_map(service)
+            )
+
+        info = find_component(
+            service["components"], route=route, usage=usage, privatelink_connection_id=privatelink_connection_id
+        )
+        host = info["host"]
+        port = info["port"]
+        if username == "default":
```
dmitrii-vasilev, line 25: What if I want to have `default` username?

aiven/client/cli.py
```diff
@@ -1488,9 +1511,11 @@ def service__cli(self):
             command, params, env = self._build_influx_start_info(url)
         elif service_type == "postgres":
             command, params, env = self._build_psql_start_info(url)
+        elif service_type == "rediss":
```
dmitrii-vasilev, line 1514: Maybe we should support `redis` as well? It seems some projects use `redis` instead of `rediss`.

dmitrii-vasilev (changes requested):
A couple of comments. Otherwise looks good.

aiven/client/connection_info/redis.py
ngilles-aiven: This is just a Redis specificity from Redis version < 6 not actually having users, and Redis >= 6 having implicit user called "default" when authenticating without a username.

ngilles-aiven: This is just a Redis specificity from Redis version < 6 not actually having users, and Redis >= 6 having implicit user called "default" when authenticating without a username.

aiven/client/cli.py
ngilles-aiven: I don't think we're using bare `redis` (i.e. non TLS wrapped) connection in Aiven platform, so this should be fine

dmitrii-vasilev (approved):
LGTM

PR #276, opened 2022-02-22
Add type annotations for aiven.client.client.py

dangquangdon:


aiven/client/client.py
```diff
@@ -1217,13 +1248,13 @@ def request_project_vpc_peering_connection(self, project, project_vpc_id, peer_c
...
-        peer_vpc,
-        peer_region=None,
-        peer_resource_group=UNDEFINED,
-    ):
+        project: str,
+        project_vpc_id: str,
+        peer_cloud_account: str,
+        peer_vpc: str,
+        peer_region: Optional[str] = None,
+        peer_resource_group: Any = UNDEFINED,
```
rominf, line 1256: `Any` -> `Union[object, str]`
Makes no difference for mypy, but preserves an info about `str`.

aiven/client/client.py
```diff
@@ -1274,7 +1305,9 @@ def get_project_vpc_peering_connection(
         )
         raise KeyError(msg)
 
-    def update_project_vpc_user_peer_network_cidrs(self, project, project_vpc_id, add=None, delete=None):
+    def update_project_vpc_user_peer_network_cidrs(
+        self, project: str, project_vpc_id: str, add: Optional[List[str]] = None, delete: Optional[List[str]] = None
```
rominf, line 1309: `add: Optional[List[str]]` -> `add: Optional[Mapping[str, str]]`

aiven/client/client.py
```diff
@@ -1306,7 +1339,7 @@ def update_project_tags(self, project, tag_updates):
         body = {"tags": tag_updates}
         return self.verify(self.patch, path, body=body)
 
-    def replace_project_tags(self, project, tags):
+    def replace_project_tags(self, project: str, tags: Dict) -> Dict[Any, Any]:
```
rominf, line 1342: `tags: Mapping[str, str]`

aiven/client/client.py
```diff
@@ -1317,20 +1350,20 @@ def replace_project_tags(self, project, tags):
...
-        project_vpc_id=UNDEFINED,
-        service_integrations=None,
-        termination_protection=False,
-        static_ips=(),
-    ):
+        project: str,
+        service: str,
+        service_type: str,
+        plan: str,
+        disk_space_mb: Optional[str] = None,
```
rominf, line 1357: `Optional[int]`

aiven/client/client.py
```diff
@@ -1317,20 +1350,20 @@ def replace_project_tags(self, project, tags):
...
+        cloud: Optional[str] = None,
+        user_config: Optional[Dict] = None,
```
rominf, line 1359: `Optional[Mapping[str, Any]]`

aiven/client/client.py
```diff
@@ -1317,20 +1350,20 @@ def replace_project_tags(self, project, tags):
...
+        project_vpc_id: object = UNDEFINED,
```
rominf, line 1360: `Union[object, str]`

aiven/client/client.py
```diff
@@ -1317,20 +1350,20 @@ def replace_project_tags(self, project, tags):
...
+        service_integrations: Optional[List[Dict]] = None,
```
rominf, line 1361: `Optional[Sequence[Mapping[str, str]]]`

aiven/client/client.py
```diff
@@ -1317,20 +1350,20 @@ def replace_project_tags(self, project, tags):
...
+        termination_protection: bool = False,
+        static_ips: Tuple = (),
```
rominf, line 1363: `Tuple[str]`

aiven/client/client.py
```diff
@@ -101,27 +103,35 @@ def _execute(self, func, method, path, body, params=None):
...
-    def verify(self, op, path, body=None, params=None, result_key=None, retry=None):
+    def verify(
+        self,
+        op: Callable[..., Response],
+        path: str,
+        body: Any = None,
+        params: Any = None,
+        result_key: Optional[str] = None,
+        retry: Optional[int] = None
+    ) -> Dict[Any, Any]:
```
rominf, line 134: I think it's better to return `Mapping[Any, Any]` to avoid surprises causes by mutations of the result. So, all other places which call `verify` need to be updated too.

aiven/client/client.py
```diff
@@ -288,40 +300,43 @@ def create_service_user(self, project, service, username, extra_params=None):
...
+    def reset_service_user_password(self, project: str, service: str, username: str, password: str) -> Dict[Any, Any]:
         path = self.build_path("project", project, "service", service, "user", username)
         body = {"operation": "reset-credentials"}
         if password is not None:
             body["new_password"] = password
         return self.verify(self.put, path, body=body)
 
-    def set_service_user_access_control(self, project, service, username, access_control):
+    def set_service_user_access_control(self, project: str, service: str, username: str,
+                                        access_control: Dict[str, Any]) -> Dict[Any, Any]:
```
rominf, line 324: `Dict[str, Any]` -> `Mapping[str, Any]`

rominf: Everywhere in arguments too.

aiven/client/client.py
```diff
@@ -566,18 +582,18 @@ def list_service_topics(self, project, service):
...
+        project: str,
+        service: str,
+        topic: str,
+        partitions: str,
+        replication: str,
+        min_insync_replicas: str,
+        retention_bytes: str,
+        retention_hours: str,
+        cleanup_policy: str,
+        tags: Optional[List] = None,
```
rominf, line 594: `tags: Optional[Sequence[str]]`

aiven/client/client.py
```diff
@@ -596,17 +612,17 @@ def create_service_topic(
...
-        body = {
+        project: str,
+        service: str,
+        topic: str,
+        partitions: str,
+        retention_bytes: str,
+        retention_hours: str,
+        min_insync_replicas: str,
+        replication: Optional[str] = None,
+        tags: Optional[List] = None,
```
rominf, line 623: `tags: Optional[Sequence[str]]`

aiven/client/client.py
```diff
@@ -668,14 +684,14 @@ def _del_es_acl_rules(config, user, rules):
...
-        add_rules=None,
-        del_rules=None,
-    ):
+        project: str,
+        service: str,
+        enabled: Optional[bool] = None,
+        extended_acl: Optional[bool] = None,
+        username: Optional[str] = None,
+        add_rules: Optional[List] = None,
+        del_rules: Optional[List] = None,
```
rominf, line 693: `Optional[Sequence[str]]`

rominf (changes requested):
1. Please check real type (e.g. you've used `str` instead of `int` in many places; it's better to not have type hints at all, than to have misleading type hints).
2. Please use specific types (e.g. `List` -> `Sequence[str]`).
3. Please address the comments and fix all similar places.

.gitignore
```diff
@@ -11,3 +11,4 @@ Pipfile*
 /rpms/
 __pycache__/
 .mypy_cache/
+.vscode
```
rominf, line 14: Unrelated, please put it into separate commit.

aiven/client/client.py
```diff
@@ -33,31 +35,31 @@ def __init__(self, response, status=520):
...
         self.http_log.propagate = False
         if show_http:
             self.http_log.setLevel(logging.DEBUG)
 
-    def set_auth_token(self, token):
+    def set_auth_token(self, token: str) -> None:
         self.auth_token = token
 
-    def set_ca(self, ca):
+    def set_ca(self, ca: bool) -> None:
```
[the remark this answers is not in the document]
dangquangdon: ah, thanks, not sure why I got the idea that it's bool at the first place

aiven/client/client.py
```diff
@@ -566,18 +582,18 @@ def list_service_topics(self, project, service):
...
-        retention_bytes,
-        retention_hours,
-        cleanup_policy,
-        tags=None,
-    ):
-        body = {
+        project: str,
+        service: str,
+        topic: str,
+        partitions: str,
```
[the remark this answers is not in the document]
dangquangdon: thanks. Some arguments doesn't have the `type` specified, for example in [this line](https://github.com/aiven/aiven-client/blob/08fe8c30cadaf141c559a4ae8aadf6557662f187/aiven/client/cli.py#L2614), I was assuming that it was in the default type of `str`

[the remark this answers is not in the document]
dangquangdon: ahh, nevermind, I didn't know that it was specified also in `cliarg.py`

aiven/client/client.py
```diff
@@ -153,7 +163,7 @@ def verify(self, op, path, body=None, params=None, result_key=None, retry=None):
 
         result = response.json()
         if result.get("error"):
-            raise Error(
+            raise Exception(
```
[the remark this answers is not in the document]
rominf: I think we want to distinguish between our exceptions and stdlib exceptions. Even if we want to change this, it's out of scope for this PR.

aiven/client/client.py
```diff
@@ -1234,21 +1270,21 @@ def delete_project_vpc_peering_connection(
...
-        peer_region=UNDEFINED,
-        peer_resource_group=UNDEFINED,
-    ):
+        project: str,
+        project_vpc_id: str,
+        peer_cloud_account: str,
+        peer_vpc: str,
+        peer_region: Union[object, str] = UNDEFINED,
+        peer_resource_group: Union[object, str] = UNDEFINED,
+    ) -> Optional[Dict]:
```
rominf, line 1287: Why not `Mapping`?

aiven/client/client.py
```diff
@@ -1317,20 +1359,20 @@ def replace_project_tags(self, project, tags):
...
-        termination_protection=False,
-        static_ips=(),
-    ):
+        project: str,
+        service: str,
+        service_type: str,
+        plan: str,
+        disk_space_mb: Optional[int] = None,
+        cloud: Optional[str] = None,
+        user_config: Optional[Mapping] = None,
```
rominf, line 1368: `Optional[Mapping[str, Any]]`?

aiven/client/client.py
```diff
@@ -1353,20 +1395,20 @@ def create_service(
...
-        karapace=None,
-        powered=None,
-        termination_protection=None,
-        project_vpc_id=UNDEFINED,
-    ):
+        project: str,
+        service: str,
+        cloud: Optional[str] = None,
+        maintenance: Optional[Mapping] = None,
+        user_config: Optional[Mapping] = None,
```
rominf, line 1402: `Optional[Mapping[str, Any]]`

aiven/client/client.py
```diff
@@ -1353,20 +1395,20 @@ def create_service(
...
+        plan: Optional[str] = None,
+        disk_space_mb: Optional[int] = None,
+        karapace: Optional[str] = None,
```
rominf, line 1405: `Optional[bool]`

aiven/client/client.py
```diff
@@ -1530,7 +1576,7 @@ def list_service_tags(self, project, service):
         )
         return self.verify(self.get, path)
 
-    def update_service_tags(self, project, service, tag_updates):
+    def update_service_tags(self, project: str, service: str, tag_updates: Mapping) -> Mapping:
```
rominf, line 1579: `tag_updates: Mapping[str, str]`

aiven/client/client.py
```diff
@@ -1541,7 +1587,7 @@ def update_service_tags(self, project, service, tag_updates):
         body = {"tags": tag_updates}
         return self.verify(self.patch, path, body=body)
 
-    def replace_service_tags(self, project, service, tags):
+    def replace_service_tags(self, project: str, service: str, tags: Mapping) -> Mapping:
```
rominf, line 1590: `tags: Mapping[str, str]`

aiven/client/client.py
```diff
@@ -1586,30 +1632,32 @@ def update_account(self, account_id, account_name):
             result_key="account",
         )
 
-    def get_accounts(self):
+    def get_accounts(self) -> Mapping:
         return self.verify(self.get, "/account", result_key="accounts")
 
-    def create_account_authentication_method(self, account_id, method_name, method_type, options=None):
+    def create_account_authentication_method(
+        self, account_id: str, method_name: str, method_type: str, options: Optional[Mapping] = None
```
rominf, line 1639: `options: Optional[Mapping[str, str]]`

aiven/client/client.py
```diff
@@ -1586,30 +1632,32 @@ def update_account(self, account_id, account_name):
...
-        authentication_id,
-        method_name=None,
-        method_enable=None,
-        options=None,
-    ):
+        account_id: str,
+        authentication_id: str,
+        method_name: Optional[str] = None,
+        method_enable: Optional[str] = None,
+        options: Optional[Mapping] = None,
```
rominf, line 1659: `options: Optional[Mapping[str, str]]`

rominf (changes requested):
Mostly LGTM. Thanks for your work. Some comments though.

@dmitrii-vasilev What do you think about returning `Mapping` everywhere? On the one hand, it's too generic, on the other, we don't need to know more detailed type, and if we decide to introduce more specific type, we will have to use `cast`...

aiven/client/client.py
dangquangdon: from `AivenCLI._tag_update_body_from_args`, it seems that it might be `None` value as well, so I guess `Mapping[str, Optional[str]]` would be more precise?

dmitrii-vasilev (approved):
LGTM

PR #277, opened 2022-03-04
Add support for new Flink table creation API

jjaakola-aiven:

This PR adds the support of new table creation API for Aiven for Apache Flink service.
Command line has new parameter `--table-properties` which takes a JSON string or a path to JSON file.

Old API is removed.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -3220,43 +3220,54 @@ ssl.truststore.type=JKS
 
     @arg.project
     @arg.service_name
-    @arg("integration_id", help="Service integration ID")
-    @arg("-n", "--table-name", required=True, help="Table name")
-    @arg("--kafka-topic", required=False, help="Topic name, used as a source/sink. (Kafka integration only)")
+    @arg("integration_id", default=None, nargs="?", help="Service integration ID (deprecated)")
+    @arg("-n", "--table-name", required=False, help="Table name (deprecated)")
+    @arg("--kafka-topic", required=False, help="Topic name, used as a source/sink. (Kafka integration only) (deprecated)")
     @arg(
         "--kafka-connector-type",
         required=False,
-        help="Kafka connector type (Kafka integration only)",
+        help="Kafka connector type (Kafka integration only) (deprecated)",
         choices=["upsert-kafka", "kafka"],
     )
     @arg(
         "--kafka-key-format",
         required=False,
-        help="Key format. (Kafka integration only)",
+        help="Key format. (Kafka integration only) (deprecated)",
         choices=["avro", "avro-confluent", "debezium-avro-confluent", "debezium-json", "json"],
     )
-    @arg("--kafka-key-fields", nargs="*", default=[], required=False, help="Key fields names used in table schema")
+    @arg(
+        "--kafka-key-fields",
+        nargs="*",
+        default=[],
+        required=False,
+        help="Key fields names used in table schema (Kafka integration only) (deprecated)"
+    )
     @arg(
         "--kafka-value-format",
         required=False,
-        help="Value format. (Kafka integration only)",
+        help="Value format. (Kafka integration only) (deprecated)",
         choices=["avro", "avro-confluent", "debezium-avro-confluent", "debezium-json", "json"],
     )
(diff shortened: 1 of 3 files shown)


juha-aiven (dismissed):
`make validate-style` is failing.

Why not remove the old API?

jjaakola-aiven:
> `make validate-style` is failing.


Fails in code that is not touched. I'll add a cleanup commit.
 
> Why not remove the old API?
Later PR, this is for having the support for the new API.

aiven/client/cliarg.py
```diff
@@ -43,6 +43,33 @@ def wrapped(self):
...
+
+    def wrapper(fun):
+        arg(
+            param_name,
+            help="JSON string or path (preceded by '@') to a JSON configuration file",
+        )(fun)
+
+        @wraps(fun)
+        def wrapped(self):
+            sanitized_param_name = param_name.replace("--", "").replace("-", "_")
```
juha-aiven, line 57: Why is this `replace("-", "_")` needed? For example if the param name in the request is `schema_sql`, is the idea here that `schema-sql` gets converted into `schema_sql`? Why wouldn't the caller have `schema_sql` in the first place. That's the name of the parameter that can be seen in the API docs.

What I think that could be the idea is that before this PR the switch to use is `--schema-sql` and thus it's kind of logical to support `schema-sql` in the JSON.

But I don't this should be done, since the name of the parameter in the API is not `schema-sql`.

And why `replace("--", "")`?

aiven/client/cli.py
```diff
@@ -3144,43 +3144,54 @@ def service__flink__table__list(self):
...
     @arg(
         "-l",
         "--like-options",
         required=False,
-        help="Clause can be used to create a table based on a definition of an existing table"
+        help="Clause can be used to create a table based on a definition of an existing table (deprecated)"
     )
-    @arg("-s", "--schema-sql", required=True, help="Source/Sink table schema")
+    @arg("-s", "--schema-sql", required=False, help="Source/Sink table schema (deprecated)")
+    @arg.json_path_or_string_parameter("--table-properties")
```
juha-aiven, line 3194: Mo hint of the format for the JSON? Should there be - it's pretty complex.

And no client side validation?

juha-aiven (changes requested):
Left few comments.

aiven/client/cliarg.py
jjaakola-aiven: The `-` to `_` is replaced as it will be the name of the member in the `self.args`.

The replace for `--` is the same as in the `user_config_json` but with different flavor as the parameter name is known there. In here the parameter name is not known and sanitazion is done programmatically.
See here: https://github.com/aiven/aiven-client/blob/a99d7eabe0183e2544aa85d3aa104de3dc1c1856/aiven/client/cliarg.py#L89-L92

What could be done is to move the logic when calling `arg` on line 50. Or change have specialized function for `--table-properties` like what is done with `--user-config-json`.

aiven/client/cli.py
jjaakola-aiven: Nothing apart of valid JSON check.

juha-aiven (approved):
LGTM

aiven/client/client.py
```diff
@@ -1070,18 +1070,7 @@ def create_flink_table(
...
-        kafka_topic: Optional[str] = None,
-        kafka_connector_type: Optional[str] = None,
-        kafka_key_format: Optional[str] = None,
-        kafka_key_fields: Optional[str] = None,
-        kafka_value_format: Optional[str] = None,
-        kafka_startup_mode: Optional[str] = None,
-        jdbc_table: Optional[str] = None,
-        opensearch_index: Optional[str] = None,
-        like_options: Optional[str] = None
+        table_properties: Dict[str, str],
```
rominf, line 1073: Since you're don't modify it, it's better to make it `Mapping`.

rominf (changes requested):

juha-aiven (approved):
Looks good.

This is a breaking change in the sense that existing functionality is removed - "Old API". This needs to be mentioned in the release notes.

PR #278, opened 2022-03-15
Add type annotations for aiven.client.cliarg and aiven.client.connection_info

rominf:

### DIFF
--- a/aiven/client/cliarg.py
+++ b/aiven/client/cliarg.py
@@ -2,15 +2,16 @@
 #
 # This file is under the Apache License, Version 2.0.
 # See the file `LICENSE` for details.
-
-from .argx import arg, UserError
+# type: ignore[attr-defined]
+from .argx import arg, CommandLineTool, UserError
 from functools import wraps
+from typing import Any, Callable, Dict, TypeVar
 
 import json as jsonlib
 import os
 
 
-def get_json_config(path_or_string):
+def get_json_config(path_or_string: str) -> Dict[str, Any]:
     # If parameter is empty, return an empty dict
     if not path_or_string:
         return {}
@@ -22,15 +23,18 @@ def get_json_config(path_or_string):
     return jsonlib.loads(path_or_string)
 
 
-def json_path_or_string(param_name):
-    def wrapper(fun):
+T = TypeVar("T")
+
+
+def json_path_or_string(param_name: str) -> Callable[[Callable[[CommandLineTool], T]], Callable[[CommandLineTool], T]]:
+    def wrapper(fun: Callable[[CommandLineTool], T]) -> Callable[[CommandLineTool], T]:
         arg(
             param_name,
             help="JSON string or path (preceded by '@') to a JSON configuration file",
         )(fun)
 
         @wraps(fun)
-        def wrapped(self):
+        def wrapped(self: CommandLineTool) -> T:
             setattr(
                 self.args,
                 param_name,
@@ -43,10 +47,10 @@ def json_path_or_string(param_name):
     return wrapper
 
 
-def user_config_json():
+def user_config_json() -> Callable[[Callable[[CommandLineTool], T]], Callable[[CommandLineTool], T]]:
     """User config that accepts arbitrary JSON"""
 
-    def wrapper(fun):
+    def wrapper(fun: Callable[[CommandLineTool], T]) -> Callable[[CommandLineTool], T]:
         arg(
             "--user-config-json",
             default=None,
@@ -55,8 +59,9 @@ def user_config_json():
         )(fun)
 
         @wraps(fun)
-        def wrapped(self):
-            if ("user_config" in self.args and (self.args.user_config_json and self.args.user_config)):
+        def wrapped(self: CommandLineTool) -> T:
+            assert self.args is not None
+            if "user_config" in self.args and (self.args.user_config_json and self.args.user_config):
                 raise UserError("-c (user config) and --user-config-json parameters can not be used at the same time")
             try:
                 setattr(
--- a/aiven/client/connection_info/_utils.py
+++ b/aiven/client/connection_info/_utils.py
@@ -1,13 +1,15 @@
 # Copyright (c) 2021 Aiven, Helsinki, Finland. https://aiven.io/
 from .common import ConnectionInfoError
 from aiven.client.common import UNDEFINED
-from typing import Mapping, Optional
+from typing import Any, Mapping, Optional, Sequence, TypeVar
 
 import ipaddress
 import urllib.parse
 
+T = TypeVar("T")
 
-def find_component(items, **filters):
+
+def find_component(items: Sequence[Mapping[str, T]], **filters: str) -> Mapping[str, T]:
     for item in items:
         if all(key in item and item[key] == value for key, value in filters.items() if value is not UNDEFINED):
             return item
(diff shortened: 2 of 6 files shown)


aiven/client/cliarg.py
```diff
@@ -2,15 +2,16 @@
 #
 # This file is under the Apache License, Version 2.0.
 # See the file `LICENSE` for details.
-
-from .argx import arg, UserError
+# type: ignore[attr-defined]
```
dmitrii-vasilev, line 5: Is it hard to avoid this ignore?

I guess it's mostly about `self.args`? Can we define `self.args` type as a protocol with `__setattribute__` like [here](https://github.com/aiven/aiven-core/blob/master/aiven/common/daemon.py#L25) for example?

aiven/client/cliarg.py
```diff
@@ -2,15 +2,16 @@
...
+from .argx import arg, CommandLineTool, UserError
 from functools import wraps
+from typing import Any, Callable
 
 import json as jsonlib
 import os
 
 
-def get_json_config(path_or_string):
+def get_json_config(path_or_string: str) -> Any:
```
dmitrii-vasilev, line 14: Shouldn't the result be restricted to `Dict` (or maybe even `Dict[str, Any]`) instead?

aiven/client/cliarg.py
```diff
@@ -22,15 +23,15 @@ def get_json_config(path_or_string):
     return jsonlib.loads(path_or_string)
 
 
-def json_path_or_string(param_name):
-    def wrapper(fun):
+def json_path_or_string(param_name: str) -> Callable[[Callable[[CommandLineTool], Any]], Callable[[CommandLineTool], Any]]:
```
dmitrii-vasilev, line 26: I'd use a type variable instead of `Any` for the result here and for all internal functions.

aiven/client/cliarg.py
```diff
@@ -43,10 +44,10 @@ def wrapped(self):
     return wrapper
 
 
-def user_config_json():
+def user_config_json() -> Callable[[Callable[[CommandLineTool], Any]], Callable[[CommandLineTool], Any]]:
```
dmitrii-vasilev, line 47: A type variable instead of `Any` for the result

aiven/client/connection_info/_utils.py
```diff
@@ -1,21 +1,21 @@
...
 from aiven.client.common import UNDEFINED
-from typing import Mapping, Optional
+from typing import Any, Mapping, Optional, Sequence
 
 import ipaddress
 import urllib.parse
 
 
-def find_component(items, **filters):
+def find_component(items: Sequence[Mapping[str, Any]], **filters: str) -> Mapping[str, Any]:
```
dmitrii-vasilev, line 10: A type variable instead of `Any`?

aiven/client/connection_info/_utils.py
```diff
@@ -1,21 +1,21 @@
...
     for item in items:
         if all(key in item and item[key] == value for key, value in filters.items() if value is not UNDEFINED):
             return item
     msg_filters = ", ".join(f"{field}={value}" for field, value in filters.items())
     raise ConnectionInfoError(f"Could not find connection information with filters {msg_filters}")
 
 
-def find_user(service, username):
+def find_user(service: Mapping[str, Any], username: str) -> Mapping[str, Any]:
```
dmitrii-vasilev, line 18: A type variable instead of `Any`?

aiven/client/connection_info/common.py
```diff
@@ -19,7 +20,7 @@ class Store(Enum):
     write = object()
     skip = object()
 
-    def handle(self, getter, path):
+    def handle(self, getter: Callable[..., str], path: str) -> None:
```
dmitrii-vasilev, line 23: Looks like it should be `Callable[[], str]` instead. The `...` means "any arguments" but the getter is used without any arguments.

aiven/client/connection_info/kafka.py
```diff
@@ -1,30 +1,32 @@
...
-    ):
-        self.host = host
-        self.port = port
-
-    def _kafkacat(self, protocol, ca_path, extra, store: Store, get_project_ca):
+    host: str
+    port: int
+
+    def _kafkacat(self, protocol: str, ca_path: str, extra: Sequence[str], store: Store,
+                  get_project_ca: Callable[..., str]) -> Sequence[str]:
```
dmitrii-vasilev, line 16: Looks like `Callable[[], str]` again.

aiven/client/connection_info/kafka.py
```diff
@@ -1,30 +1,32 @@
...
-    def __init__(self, host, port, client_cert, client_key):
-        super().__init__(host, port)
-        self.client_cert = client_cert
-        self.client_key = client_key
+    client_cert: str
+    client_key: str
 
-    def kafkacat(self, store, get_project_ca, ca_path, client_key_path, client_cert_path):
+    def kafkacat(
+        self, store: Store, get_project_ca: Callable[..., str], ca_path: str, client_key_path: str, client_cert_path: str
```
dmitrii-vasilev, line 28: `Callable[[], str]`

aiven/client/connection_info/kafka.py
```diff
@@ -98,7 +99,7 @@ def from_service(
             raise ConnectionInfoError(f"Could not find password for username {username}")
         return cls(host=info["host"], port=info["port"], username=username, password=user["password"])
 
-    def kafkacat(self, store, get_project_ca, ca_path):
+    def kafkacat(self, store: Store, get_project_ca: Callable[..., str], ca_path: str) -> Sequence[str]:
```
dmitrii-vasilev, line 102: Probably `Callable[[], str]` again?

mypy.ini
```diff
@@ -34,22 +34,7 @@ ignore_missing_imports = True
...
-
-[mypy-aiven.client.connection_info.common]
-ignore_errors = True
-
-[mypy-aiven.client.connection_info.kafka]
-ignore_errors = True
-
-[mypy-aiven.client.connection_info.pg]
-ignore_errors = True
+warn_unused_ignores = False
```
dmitrii-vasilev, line 37: Can it be avoided?

dmitrii-vasilev (changes requested):
Some comments

aiven/client/connection_info/_utils.py
```diff
@@ -1,22 +1,27 @@
...
     for item in items:
         if all(key in item and item[key] == value for key, value in filters.items() if value is not UNDEFINED):
             return item
     msg_filters = ", ".join(f"{field}={value}" for field, value in filters.items())
     raise ConnectionInfoError(f"Could not find connection information with filters {msg_filters}")
 
 
-def find_user(service, username):
-    for user in service["users"]:
+def find_user(service: Mapping[str, T], username: str) -> Mapping[str, T]:
```
dmitrii-vasilev, line 21: I think type variable is wrong and confusing here. Type for `service` argument looks like `Mapping[str, Mapping[str, Mapping[str, str]]]` and the result is probably `Mapping[str, str]`. So if type for values in the result isn't always `str` then probably `service` argument should be `Mapping[str, Mapping[str, Mapping[str, T]]]` instead.

dmitrii-vasilev (dismissed):
It seems type variable is wrong in this case

aiven/client/connection_info/_utils.py
rominf: I think should not assume more than this:
```
def find_user(service: Mapping[str, Any], username: str) -> Mapping[str, Any]:
    users = service["users"]
    assert isinstance(users, Mapping)
```
since it's arbitrary depth JSON which can contain integers, nulls, etc.

dmitrii-vasilev (approved):
LGTM

PR #279, opened 2022-03-16
Add type annotations to `aiven.client.speller` and `tests/test_speller.py`

mlatief:
Add type annotations to:

- [x] `aiven/client/speller.py`
- [x] `tests/test_speller.py`

Part of https://github.com/aiven/aiven-client/issues/258

### DIFF
--- a/aiven/client/speller.py
+++ b/aiven/client/speller.py
@@ -3,22 +3,24 @@
 # Copyright (c) 2007-2016 Peter Norvig
 # MIT license: www.opensource.org/licenses/mit-license.php
 
+from typing import Container, Iterable, Iterator, Optional
 
-def suggest(word_to_check, known_words):
-    def get_correction(word):
+
+def suggest(word_to_check: str, known_words: Container[str]) -> Optional[str]:
+    def get_correction(word: str) -> Optional[str]:
         """Most probable spelling correction for word."""
         candidates = get_candidates(word)
         return next(iter(candidates)) if candidates else None
 
-    def get_candidates(word):
+    def get_candidates(word: str) -> Iterable[str]:
         """Generate possible spelling corrections for word."""
         return get_known([word]) or get_known(get_edits1(word)) or get_known(get_edits2(word)) or []
 
-    def get_known(words):
+    def get_known(words: Iterable[str]) -> Iterable[str]:
         """The subset of `words` that appear in the dictionary of WORDS."""
         return set(w for w in words if w in known_words)
 
-    def get_edits1(word):
+    def get_edits1(word: str) -> Iterable[str]:
         """All edits that are one edit away from `word`."""
         letters = "abcdefghijklmnopqrstuvwxyz_"
         splits = [(word[:i], word[i:]) for i in range(len(word) + 1)]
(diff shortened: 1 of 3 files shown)


aiven/client/speller.py
```diff
@@ -4,21 +4,24 @@
...
+
+
+def suggest(word_to_check: str, known_words: Iterable[str]) -> Optional[str]:
+    def get_correction(word: str) -> Optional[str] :
         """Most probable spelling correction for word."""
         candidates = get_candidates(word)
         return next(iter(candidates)) if candidates else None
 
-    def get_candidates(word):
+    def get_candidates(word: str) -> Iterable[str]:
```
dmitrii-vasilev, line 16: Probably can be `Iterator[str]` instead. The result is only used once in `get_correction()`

aiven/client/speller.py
```diff
@@ -4,21 +4,24 @@
...
         """Generate possible spelling corrections for word."""
         return get_known([word]) or get_known(get_edits1(word)) or get_known(get_edits2(word)) or []
 
-    def get_known(words):
+    def get_known(words: Iterable[str]) -> Iterable[str]:
```
dmitrii-vasilev, line 20: Also, probably both types can be `Iterator[str]` instead

aiven/client/speller.py
```diff
@@ -4,21 +4,24 @@
...
         """The subset of `words` that appear in the dictionary of WORDS."""
         return set(w for w in words if w in known_words)
 
-    def get_edits1(word):
+    def get_edits1(word: str) -> Iterable[str]:
```
dmitrii-vasilev, line 24: Consider `Iterator[str]`

aiven/client/speller.py
```diff
@@ -28,7 +31,7 @@ def get_edits1(word):
         inserts = [L + c + R for L, R in splits for c in letters]
         return set(deletes + transposes + replaces + inserts)
 
-    def get_edits2(word):
+    def get_edits2(word: str) -> Iterable[str]:
```
dmitrii-vasilev, line 34: Here it should be `Iterator[str]` because it returns iterator comprehension.

tests/test_speller.py
```diff
@@ -37,5 +38,5 @@
         ("asdf", SERVICE_TYPES, None),
     ],
 )
-def test_suggest(word_to_check, known_words, suggestion):
+def test_suggest(word_to_check: str, known_words: List[str], suggestion: Optional[str]) -> None:
```
dmitrii-vasilev, line 41: Probably `known_words: Container[str]`

dmitrii-vasilev (changes requested):
I think we can tighten some types here.

aiven/client/speller.py
mlatief: I think considering the return value I will keep this one as `Iterable`.

aiven/client/speller.py
mlatief: Also here considering the return value I will keep it as `Iterable`.

dmitrii-vasilev (approved):
LGTM

PR #280, opened 2022-03-18
Use Black code formatter

rominf:

Black code formatter is faster, more popular, and is used in other Aiven projects.


kmichel-aiven (approved):

mlatief (approved):
Looks awesome to me.

PR #281, opened 2022-03-21
Add type annotations to aiven.client.pretty and tests/test_pretty.py

mlatief:
Add type annotations to:

- [x] `aiven/client/pretty.py`
- [x] `tests/test_pretty.py`

Part of https://github.com/aiven/aiven-client/issues/258


aiven/client/pretty.py
```diff
@@ -178,8 +172,22 @@ def iter_values(key, value):
 
 
 def print_table(
-    result, drop_fields: Optional[Collection[str]] = None, table_layout: Optional[TableLayout] = None, header=True, file=None
-):  # pylint: disable=redefined-builtin
+    result: Optional[Union[Collection[Any], ResultType]],
+    drop_fields: Optional[Collection[str]] = None,
+    table_layout: Optional[TableLayout] = None,
+    header: bool = True,
+    file: Optional[IO] = None,
```
rominf, line 179: I think it should be `Optional[TextIO]`.

rominf (commented):
LGTM. Just one comment.

rominf:
As this module is quite complicated, I would propose @dmitrii-vasilev check it too.

rominf (approved):

aiven/client/pretty.py
```diff
@@ -19,11 +20,12 @@
 except NameError:
     basestring = str  # pylint: disable=redefined-builtin
 
+ResultType = Collection[Mapping[str, Any]]
 TableLayout = Collection[Union[List[str], Tuple[str], str]]
 
 
 class CustomJsonEncoder(json.JSONEncoder):
-    def default(self, o):  # pylint:disable=E0202
+    def default(self, o: Any) -> Any:  # pylint:disable=E0202
```
dmitrii-vasilev, line 28: Hmm, shouldn't the result type be `str`?

aiven/client/pretty.py
```diff
@@ -161,8 +154,9 @@ def iter_values(key, value):
         # The main, horizontal, line
         yield "  ".join(formatted_row.get(f, "").ljust(widths[f]) for f in horizontal_fields).strip()
         # And the rest of the fields, one per field
-        fields_to_print = []
-        for vertical_field in table_layout[1:]:
+        fields_to_print: List[Tuple[str, str]] = []
+        vertical_fields = cast(Iterator[str], itertools.islice(table_layout, 1, None))
```
dmitrii-vasilev, line 158: Is the `cast` really needed here?

dmitrii-vasilev (changes requested):
A couple of small comments. Looks good otherwise

aiven/client/pretty.py
mlatief: Without it, mypy can't narrow the iterable to just `Iterable[str]`, especially given how `table_layout` is annotated. When used in the for loop without casting, `vertical_field` would just be inferred as `List[str] | Tuple[str] | str`, yet it is assumed that all values aside from the first one in `table_layout`, are always `str` representing the vertical fields, if any.

dmitrii-vasilev: I was wondering if `vertical_fields: Iterator[str]` would work

dmitrii-vasilev (approved):
LGTM

aiven/client/pretty.py
mlatief: causes a conflict with `islice` generic types, seems to expect the first `__init__` argument to be of the same base type:
 ```
 aiven/client/pretty.py:158: error: Argument 1 to "islice" has incompatible type "Collection[Union[List[str], Tuple[str], str]]"; expected "Iterable[str]"
 ```

PR #282, opened 2022-03-24
Add initial support for creating OAuth2 apps

m-novikov:

Adds ability to configure OAuth2 apps for users.
As we are moving to allow third-party providers to integrate with aiven services it's helpful to have convenient interface to configure oauth2 apps.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -5115,6 +5115,42 @@ server_encryption_options:
         )
         print(response["message"])
 
+    @arg.json
+    @arg.account_id
+    @arg("-n", "--name", required=True, help="OAuth2 application name")
+    @arg(
+        "--redirect-uri",
+        required=True,
+        help="Redirect URI",
+    )
+    @arg(
+        "-d",
+        "--description",
+        required=False,
+        help="App description",
+    )
+    def account__oauth2_client__create(self):
+        """Create new OAuth2 client."""
+        oauth2_client = self.client.create_oauth2_client(
+            self.args.account_id,
+            name=self.args.name,
+            description=self.args.description,
+        )
+        client_id = oauth2_client["client_id"]
+
+        try:
+            redirect = self.client.create_oauth2_client_redirect(self.args.account_id, client_id, self.args.redirect_uri)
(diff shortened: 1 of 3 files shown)


aiven/client/client.py
```diff
@@ -2140,3 +2140,20 @@ def detach_team_from_project(self, account_id: str, team_id: str, project: str)
...
+        )
+
+    def add_oauth2_client_redirect(self, account_id: str, client_id: str, uri: str) -> Mapping:
+        return self.verify(
+            self.post,
+            self.build_path("account", account_id, "oauth_client", client_id, "redirect"),
+            body={"redirect_uri": uri},
+        )
+
+    def add_oauth2_client_secret(self, account_id: str, client_id: str) -> Mapping:
```
sjamgade, line 2158: shouldnt this be `get_oauth2....` ?

aiven/client/cli.py
```diff
@@ -5115,6 +5115,36 @@ def service__tags__replace(self):
...
+    )
+    @arg(
+        "-d",
+        "--description",
+        required=False,
+        help="App description",
+    )
+    def account__oauth2_client__create(self):
+        """Create new OAuth2 client."""
+        oauth2_client = self.client.create_oauth2_client(
```
sjamgade, line 5134: if something blows up down the line, the client will persist. 
Either remove the client then and there or allow it to be removed later ?

sjamgade (changes requested):
LGTM, some questions

aiven/client/client.py
m-novikov: Maybe create? It does post request and issues new one.

aiven/client/cli.py
m-novikov: Added removal on exception

sjamgade (approved):
LGTM

sjamgade:
@tvainika did you want to take a look at this ?
@aiven/aiven-client  maybe too ?

rominf (approved):
I'm not super familiar with this topic, but from Aiven Client team member POV, looks fine.

PR #284, opened 2022-04-04
Harden the GitHub Workflows [SECDEV-678]

st3fan:
This patch improves security around GitHub Workflows:

For the `build.yml` workflow:

 * Set the default workflow permissions to `read` - because this workflow does not need to modify anything in the project.
 * Do not let `actions/checkout` persist (write to disk) the GitHub token - because this workflow does not need to modify anything in the project. And prevents other actions to look at the token.

For the `publish-pypi.yaml` workflow:

 * Set the default workflow permissions to `read` - because this workflow does not need to modify anything in the project.
 * Do not let `actions/checkout` persist (write to disk) the GitHub token - because this workflow does not need to modify anything in the project. And prevents other actions to look at the token.
 * Pin the [`pypa/gh-action-pypi-publish`](https://github.com/pypa/gh-action-pypi-publish) action to the commit hash of their last release (`v1.5`) instead of pulling in `master`.

The other workflows only use trusted (GitHub) actions - no changes were made.

Additional notes in SECDEV-678

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,5 +1,7 @@
 name: Build Aiven Client
 
+permissions: read-all
+
 on:
   push:
     branches:
@@ -21,6 +23,8 @@ jobs:
 
       - id: checkout-code
         uses: actions/checkout@v2
+        with:
+          persist-credentials: false
 
       - id: prepare-python
         uses: actions/setup-python@v2
@@ -55,6 +59,8 @@ jobs:
     steps:
       - id: checkout-code
         uses: actions/checkout@v2
+        with:
+          persist-credentials: false
 
       - id: prepare-python
         uses: actions/setup-python@v2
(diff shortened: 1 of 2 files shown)


st3fan:
The build for this PR is failing in something unrelated to my changes. It looks like this could be related to `click` got a major bump, going from `8.0.4` to `8.1.2`. Not sure what a good next step is to resolve this @aiven/aiven-client

st3fan:
> The build for this PR is failing in something unrelated to my changes. It looks like this could be related to `click` got a major bump, going from `8.0.4` to `8.1.2`. Not sure what a good next step is to resolve this @aiven/aiven-client

Rebasing helped. Ready for review.

rominf (approved):
LGTM. Could somebody from security team review it too, since I'm not an expert here?

paulwouters (approved):
LGTM

PR #285, opened 2022-04-07
Show all results for static ips list

dangquangdon:
[DARK-637]

[DARK-637]: https://aiven.atlassian.net/browse/DARK-637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aiven/client/client.py
+++ b/aiven/client/client.py
@@ -1498,7 +1498,7 @@ class AivenClient(AivenClientBase):
 
     def list_static_ip_addresses(self, project: str) -> Mapping:
         path = self._static_ip_address_path(project)
-        return self.verify(self.get, path, result_key="static_ips")
+        return self.verify(self.get, path, result_key="static_ips", params={"limit": 999})
 
     def create_static_ip_address(self, project: str, cloud_name: str) -> Mapping:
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -1187,11 +1187,35 @@ def service__privatelink__availability(self):
 
     @arg.project
     @arg("--format", help="Format string for output")
+    @arg("--limit", help="Limit number of results showing in each page.", type=int)
+    @arg("--page", help="Page number.", type=int, default=0)
```
dmitrii-vasilev, line 1191: Hmm, I'd say that we shouldn't implement our own paging support here but just output everything. If somebody need to see the result page by page they can pipe it to an external pager like `more` or `less`.

Moreover those `PAGE n/m` headers currently break automatic parsing of the result.

dmitrii-vasilev (changes requested):
I think we shouldn't implement our own paging support

dmitrii-vasilev (approved):
LGTM

PR #286, opened 2022-04-12
Add type annotations to aiven cli

dangquangdon:
Add type annotations to `cli.py` and fix annotations in some other parts

[DARK-556]

[DARK-556]: https://aiven.atlassian.net/browse/DARK-556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aiven/client/cli.py
```diff
@@ -3737,7 +3758,7 @@ def vpc__peering_connection__delete(self):
             print(ex.response.text)
             raise
 
-    def _get_service_project_vpc_id(self):
+    def _get_service_project_vpc_id(self) -> Optional[Union[object, str]]:
```
dmitrii-vasilev, line 3761: Probably can be `Union[str, None, Literal[UNDEFINED]]` instead.

aiven/client/cli.py
```diff
@@ -3847,7 +3868,7 @@ def service__create(self):
         project_vpc_id = self._get_service_project_vpc_id()
         project = self.get_project()
         user_config_schema = self._get_service_type_user_config_schema(project=project, service_type=service_type)
-        user_config = self.create_user_config(user_config_schema)
+        user_config = {**self.create_user_config(user_config_schema)}
```
dmitrii-vasilev, line 3871: `.copy()` method or change the result type

aiven/client/cli.py
```diff
@@ -3901,15 +3922,15 @@ def _get_powered(self):
             return None
 
     @staticmethod
-    def _get_unknown_option_error(option_type, options, option):
+    def _get_unknown_option_error(option_type: str, options: Mapping, option: str) -> str:
```
dmitrii-vasilev, line 3925: Probably `Mapping[str, Any]`

aiven/client/cli.py
```diff
@@ -3901,15 +3922,15 @@ def _get_powered(self):
...
         suggestion = suggest(option, options)
         if suggestion is not None:
             did_you_mean = ", did you mean: {}".format(suggestion)
         else:
             did_you_mean = ""
         return "Unknown {} {!r}{} (available options: {})".format(option_type, option, did_you_mean, ", ".join(options))
 
-    def _get_service_type_user_config_schema(self, project, service_type):
+    def _get_service_type_user_config_schema(self, project: str, service_type: str) -> Mapping:
```
dmitrii-vasilev, line 3933: Probably `Mapping[str, Any]`

aiven/client/cli.py
```diff
@@ -3924,7 +3945,7 @@ def _get_service_type_user_config_schema(self, project, service_type):
 
         return service_def["user_config_schema"]
 
-    def _get_service_version_info(self, service_type, version):
+    def _get_service_version_info(self, service_type: str, version: str) -> Mapping:
```
dmitrii-vasilev, line 3948: Probably `Mapping[str, Any]`

aiven/client/cli.py
```diff
@@ -3980,7 +4001,7 @@ def _do_version_eol_check(self, service_type: str, requested_version: str) -> No
             if not self.confirm("continue anyway (y/N)? "):
                 raise argx.UserError("Aborted")
 
-    def _get_endpoint_user_config_schema(self, project, endpoint_type_name=None):
+    def _get_endpoint_user_config_schema(self, project: str, endpoint_type_name: str) -> Mapping:
```
dmitrii-vasilev, line 4004: `Mapping[str, Any]`

aiven/client/cli.py
```diff
@@ -3994,7 +4015,7 @@ def _get_endpoint_user_config_schema(self, project, endpoint_type_name=None):
                 )
             ) from ex
 
-    def _get_integration_user_config_schema(self, project, integration_type_name):
+    def _get_integration_user_config_schema(self, project: str, integration_type_name: str) -> Mapping:
```
dmitrii-vasilev, line 4018: `Mapping[str, Any]`

aiven/client/cli.py
```diff
@@ -4067,15 +4088,15 @@ def _get_integration_user_config_schema(self, project, integration_type_name):
...
+    def service__update(self) -> None:
         """Update service settings"""
         powered = self._get_powered()
         karapace = self.args.use_karapace
         project = self.get_project()
         service = self.client.get_service(project=project, service=self.args.service_name)
         plan = self.args.plan or service["plan"]
         user_config_schema = self._get_service_type_user_config_schema(project=project, service_type=service["service_type"])
-        user_config = self.create_user_config(user_config_schema)
+        user_config = {**self.create_user_config(user_config_schema)}
```
dmitrii-vasilev, line 4099: `.copy()` of change the result type

aiven/client/cli.py
```diff
@@ -4141,23 +4162,23 @@ def project__switch(self):
...
     @arg("project_name", help="Project name")
     @arg.cloud
-    def project__delete(self):
+    def project__delete(self) -> None:
         """Delete a project"""
         self.client.delete_project(project=self.args.project_name)
 
     @classmethod
-    def _format_card_info(cls, project):
+    def _format_card_info(cls, project: Mapping) -> str:
```
dmitrii-vasilev, line 4170: Probably `Mapping[str, Any]`

aiven/client/cli.py
```diff
@@ -4141,23 +4162,23 @@ def project__switch(self):
...
         card_info = project.get("card_info")
         if not card_info:
             return "N/A"
 
         return "{}/{}".format(project["card_info"]["user_email"], project["card_info"]["card_id"])
 
-    def _show_projects(self, projects, verbose=True):
+    def _show_projects(self, projects: Sequence[Dict], verbose: bool = True) -> None:
```
dmitrii-vasilev, line 4177: Probably `Sequence[Dict[str, Any]]`

aiven/client/cli.py
```diff
@@ -4231,23 +4252,23 @@ def project__create(self):
         self.config["default_project"] = self.args.project_name
         self.config.save()
 
-        self._show_projects([project])
+        self._show_projects([{**project}])
```
dmitrii-vasilev, line 4255:
```suggestion
        self._show_projects([project.copy()])
```

aiven/client/cli.py
```diff
@@ -4231,23 +4252,23 @@ def project__create(self):
...
 
     @arg.json
     @arg.project
-    def project__details(self):
+    def project__details(self) -> None:
         """Show project details"""
         project_name = self.get_project()
         project = self.client.get_project(project=project_name)
-        self._show_projects([project])
+        self._show_projects([{**project}])
```
dmitrii-vasilev, line 4267:
```suggestion
        self._show_projects([project.copy()])
```

aiven/client/cli.py
```diff
@@ -4288,12 +4309,12 @@ def project__update(self):
         if self.args.name and self.config["default_project"] == project_name:
             self.config["default_project"] = project["project_name"]
             self.config.save()
-        self._show_projects([project])
+        self._show_projects([{**project}])
```
dmitrii-vasilev, line 4312:
```suggestion
        self._show_projects([project.copy()])
```

aiven/client/cli.py
```diff
@@ -4333,7 +4354,7 @@ def service__keypair__get(self):
         with open(self.args.cert_filepath, "w", encoding="utf-8") as fp:
             fp.write(result["certificate"])
 
-    def _validate_service_cassandra_sstableloader(self):
+    def _validate_service_cassandra_sstableloader(self) -> Mapping:
```
dmitrii-vasilev, line 4357: Probably `Mapping[str, Any]`

aiven/client/cli.py
```diff
@@ -4570,7 +4591,7 @@ def project__invite_list(self):
             print(ex.response.text)
             raise argx.UserError("Project user listing for '{}' failed".format(project_name))
 
-    def _print_tags(self, tags):
+    def _print_tags(self, tags: Mapping) -> None:
```
dmitrii-vasilev, line 4594: `Mapping[str, Any]`

aiven/client/cli.py
```diff
@@ -4579,23 +4600,23 @@ def _print_tags(self, tags):
...
-    def _parse_tag_arg(tag):
+    def _parse_tag_arg(tag: str) -> Tuple[str, str]:
         if "=" not in tag:
             raise argx.UserError(f"Invalid tag key-value pair '{tag}', expected '<KEY>=<VALUE>'")
         key, value = tag.split("=", maxsplit=1)
         return key, value
 
-    def _tag_update_body_from_args(self):
-        tag_updates = {key: None for key in self.args.remove_tag}
+    def _tag_update_body_from_args(self) -> Mapping[str, str]:
```
dmitrii-vasilev, line 4609: Looks like `Mapping[str, Optional[str]]`

aiven/client/cli.py
```diff
@@ -4579,23 +4600,23 @@ def _print_tags(self, tags):
...
+        tag_updates: Dict = {key: None for key in self.args.remove_tag}
```
dmitrii-vasilev, line 4610: `Dict[str, Optional[str]]`

aiven/client/cli.py
```diff
@@ -4634,20 +4655,20 @@ def user__create(self):
...
-    def _write_auth_token_file(self, token, email):
+    def _write_auth_token_file(self, token: str, email: str) -> None:
         with self._open_auth_token_file(mode="w") as fp:
             fp.write(jsonlib.dumps({"auth_token": token, "user_email": email}))
             aiven_credentials_filename = fp.name
         os.chmod(aiven_credentials_filename, 0o600)
         self.log.info("Aiven credentials written to: %s", aiven_credentials_filename)
 
-    def _open_auth_token_file(self, mode="r"):
+    def _open_auth_token_file(self, mode: str = "r") -> IO:
```
dmitrii-vasilev, line 4671: Probably `TextIO`

aiven/client/cli.py
```diff
@@ -4683,7 +4704,7 @@ def _get_auth_token(self):
                 return None
             raise
 
-    def pre_run(self, func):
+    def pre_run(self, func: Callable) -> None:
```
dmitrii-vasilev, line 4707: Looks like `Callable[[], Optional[int]]`

aiven/client/cli.py
```diff
@@ -4788,18 +4811,18 @@ def credits__list(self):
...
-    def credits__claim(self):
+    def credits__claim(self) -> None:
         """Claim a credit code"""
         project_name = self.get_project()
         result = self.client.claim_project_credit(project=project_name, credit_code=self.args.code)
         if self.args.json:
             self.print_response(result, json=True)
 
-    def _print_billing_groups(self, billing_groups):
+    def _print_billing_groups(self, billing_groups: Sequence[Dict]) -> None:
```
dmitrii-vasilev, line 4821: Can be `Sequence[Dict[str, Any]]`

aiven/client/cli.py
```diff
@@ -4866,7 +4889,7 @@ def billing_group__create(self):
             state=self.args.state,
             zip_code=self.args.zip_code,
         )
-        self._print_billing_groups([billing_group])
+        self._print_billing_groups([{**billing_group}])
```
dmitrii-vasilev, line 4892:
```suggestion
        self._print_billing_groups([billing_group.copy()])
```

aiven/client/cli.py
```diff
@@ -4903,18 +4926,18 @@ def billing_group__update(self):
...
 
     @arg.billing_group
     @arg.json
     @arg.verbose
-    def billing_group__get(self):
+    def billing_group__get(self) -> None:
         """Get a project"""
         billing_group = self.client.get_billing_group(billing_group=self.args.id)
-        self._print_billing_groups([billing_group])
+        self._print_billing_groups([{**billing_group}])
```
dmitrii-vasilev, line 4940:
```suggestion
        self._print_billing_groups([billing_group.copy()])
```

aiven/client/cli.py
```diff
@@ -5129,13 +5152,15 @@ def service__tags__replace(self):
...
-            name=self.args.name,
-            description=self.args.description,
-        )
+        oauth2_client = {
+            **self.client.create_oauth2_client(
+                self.args.account_id,
+                name=self.args.name,
+                description=self.args.description,
+            )
+        }
```
dmitrii-vasilev, line 5163: `.copy()` of change the result type

dmitrii-vasilev (changes requested):
Only checked 3 files for now - `.pylintrc`, `argx.py` and `cli.py`.

aiven/client/connection_info/_utils.py
```diff
@@ -1,15 +1,15 @@
...
+from typing import Any, Mapping, Optional, Sequence, TypeVar, Union
 
 import ipaddress
 import urllib.parse
 
 T = TypeVar("T")
 
 
-def find_component(items: Sequence[Mapping[str, T]], **filters: str) -> Mapping[str, T]:
+def find_component(items: Sequence[Mapping[str, T]], **filters: Union[object, str]) -> Mapping[str, T]:
```
dmitrii-vasilev, line 12: Probably `Union[str, Literal[UNDEFINED]]`

aiven/client/connection_info/kafka.py
```diff
@@ -45,7 +45,7 @@ def from_service(
         service: Mapping[str, Any],
         *,
         route: str,
-        privatelink_connection_id: str,
+        privatelink_connection_id: Union[object, str],
```
dmitrii-vasilev, line 48: Probably `Union[str, Literal[UNDEFINED]]`

aiven/client/connection_info/redis.py
```diff
@@ -17,7 +17,14 @@ class RedisConnectionInfo:
 
     @classmethod
     def from_service(
-        cls, service: Mapping[str, Any], *, route: str, usage: str, privatelink_connection_id: str, username: str, db: str
+        cls,
+        service: Mapping[str, Any],
+        *,
+        route: str,
+        usage: str,
+        privatelink_connection_id: Union[object, str],
```
dmitrii-vasilev, line 25: Probably `Union[str, Literal[UNDEFINED]]`

dmitrii-vasilev (changes requested):
Reviewed everything except `client.py` for now.

aiven/client/argx.py
```diff
@@ -146,7 +146,7 @@ def save(self) -> None:
 class CommandLineTool:  # pylint: disable=old-style-class
     def __init__(self, name: str, parser: Optional[argparse.ArgumentParser] = None):
         self.log = logging.getLogger(name)
-        self.config: Optional[Config] = None
+        self.config: Config = cast(Config, {})
```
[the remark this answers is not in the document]
dangquangdon: it's to please `mypy` that this config is a `Config` object. Although the config is actually constructed in `CommandLine.run`, `mypy` doesn't know about it, so we need to tell that this config won't be `None` and it will be `Config` object.

aiven/client/cli.py
```diff
@@ -4,17 +4,19 @@
 # See the file `LICENSE` for details.
 from . import argx, client
 from aiven.client import envdefault
-from aiven.client.cliarg import arg
+from aiven.client.cliarg import arg  # type: ignore
```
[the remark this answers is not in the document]
dangquangdon: No, it's because `mypy` complains that `Module "aiven.client.cliarg" has no attribute "arg"`. This is perhaps because of the way we structured this lib.

aiven/client/cli.py
```diff
@@ -40,14 +42,14 @@
 EOL_ADVANCE_WARNING_TIME = timedelta(weeks=26)  # Give 6 months advance notice for EOL services
 
 
-def convert_str_to_value(schema, str_value):
-    if "string" in schema["type"]:
+def convert_str_to_value(schema: Mapping[str, Any], str_value: Optional[str]) -> Any:
+    if "string" in schema["type"] and str_value:
```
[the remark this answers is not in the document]
dangquangdon: if it's empty string, a `KeyError` or `ValueError` exception will be raised, and as I checked in the places we're using this function, we do catch the exception. And perhaps empty string is acceptable when schema type is string, not entirely sure tho.

aiven/client/cli.py
```diff
@@ -124,21 +126,26 @@ def get_current_date() -> datetime:
...
 
     class ClientFactory(Protocol):  # pylint: disable=too-few-public-methods
-        def __call__(self, base_url: str, show_http: bool, request_timeout: Optional[int]):
+        def __call__(self, base_url: str, show_http: bool, request_timeout: Optional[int]) -> client.AivenClient:
             ...
 
 
 else:
-    ClientFactory = Callable[..., client.AivenClient]
+    ClientFactory = Callable[..., client.AivenClient]  # type: ignore
```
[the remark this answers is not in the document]
dangquangdon: because `mypy` complains that `Cannot assign to a type` :thinking:

aiven/client/argx.py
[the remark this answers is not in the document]
dangquangdon: maybe it would work the same way as you suggested for the `client: AivenClient` :+1:  I'll try that

aiven/client/cli.py
```diff
@@ -489,7 +502,7 @@ def user__access_token__list(self):
         ]
         self.print_response(tokens, json=self.args.json, table_layout=layout)
 
-    def _show_logs(self, msgs):
+    def _show_logs(self, msgs: Mapping) -> str:
```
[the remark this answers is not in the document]
dangquangdon: I'll make it `Mapping[str, Any]`, because according to the response schema of service logs, the `logs` property is of type array

aiven/client/cli.py
```diff
@@ -668,15 +681,17 @@ def _parse_auth_config_options(config_cmdline, config_file):
...
-            options=options,
-        )
+        method = {
+            **self.client.create_account_authentication_method(
+                self.args.account_id,
+                method_name=self.args.name,
+                method_type=self.args.type,
+                options=options,
+            )
+        }
```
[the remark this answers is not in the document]
dangquangdon: The return type of `self.client.create_account_authentication_method` is Mapping, so it cannot be mutated. We have to convert it to a dict. I'll use the `dict()` method as you suggested.

aiven/client/cli.py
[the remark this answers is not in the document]
dmitrii-vasilev: Ah, this is because it's imported from `.argx` there. I think it was already discussed in another PR. But maybe do `arg: Type` in `cliarg` will help resolve it?

aiven/client/cli.py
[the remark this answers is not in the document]
dmitrii-vasilev: I mean the new logic is not the same as before. For example if `schema["type"]` is `string` and `str_value = ""` previously this function will return `""`. With the new changes it will raise `argx.UseError()` in the same case. Without checking all the client code it feels wrong. If this change is intended then it's OK.

In any case I think it's better to extract `and str_value` in the way I proposed in the first comment here.

aiven/client/cli.py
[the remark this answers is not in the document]
dmitrii-vasilev: I think this is because of the other branch of this `if`. I'd drop the whole `if` here and only keep this line with updated arguments of the `Callable`. I don't see a point to define it as a protocol.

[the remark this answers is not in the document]
dangquangdon: I think the protocol is to help specify and annotate the keyword arguments (once we're moving to py3.8+).  Because with `Callable` we cannot specify the name of keyword arguments that we expect. In this case if I try 
```python
Callalble[[str, bool, Optional[int]], AivenClient]
```
Then, mypy will complain about e.g `Unexpected keyword argument "base_url"` when we call `client_factory` in [`self.pre_run`](https://github.com/aiven/aiven-client/blob/371f5ead47aefc97a6df821fc45fe3a50a230f53/aiven/client/cli.py#L4640)

aiven/client/cli.py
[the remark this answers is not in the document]
dangquangdon: Generally I think when we say the return type is `Dict`, we're hinting that the return value is mutatable which is not always a good practice I guess. But for this case I think it's not that important :+1:

aiven/client/cli.py
```diff
@@ -1055,12 +1070,12 @@ def service__privatelink__aws__delete(self):
...
     @arg("--format", help="Format string for output")
     @arg.json
-    def service__privatelink__aws__connection__list(self):
+    def service__privatelink__aws__connection__list(self) -> None:
         """List PrivateLink connections for a service"""
         resp = self.client.list_service_privatelink_aws_connections(
             project=self.get_project(), service=self.args.service_name
         )
-        self.print_response([resp], format=self.args.format, json=self.args.json)
+        self.print_response(resp, format=self.args.format, json=self.args.json)
```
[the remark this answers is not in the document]
dangquangdon: because `list_service_privatelink_aws_connections` is actually returning a `Sequence`, not a `Mapping`. And in `print_response`, it's annotated that it accept `Union[Mapping[str, Any], Collection[Mapping[str, Any]]]`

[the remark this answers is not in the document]
dangquangdon: To add more information, with the old way of `[resp]` the result was printed as a dict, without proper table format like this 
```
{"dns_name": "some-user-name.aws.com", "privatelink_connection_id": "plc3abd778b14d", "state": "active", "vpc_endpoint_id": "vpce-01234567abcdef"}
```

Now we fix, it would print the table like this:
```
DNS_NAME                PRIVATELINK_CONNECTION_ID  STATE   VPC_ENDPOINT_ID    
======================  =========================  ======  ===================
some-user-name.aws.com  plc3abd778b14d             active  vpce-01234567abcdef
```

aiven/client/cli.py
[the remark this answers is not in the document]
dmitrii-vasilev: Ah, right. I'm wondering if something like the following will work?
```python
class ClientFactory(Callable[..., client.AivenClient]):
    pass
```

Otherwise let's keep `type: ignore`.

aiven/client/cli.py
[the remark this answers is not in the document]
dmitrii-vasilev: Yes, it's not a good practice. But can be useful in some cases - for example to give the caller a bunch of default values to update. I think the important restriction here that the dict should be created inside the function and not passed from some other place.

aiven/client/client.py
```diff
@@ -185,10 +185,10 @@ def build_path(*parts: str) -> str:
 class AivenClient(AivenClientBase):
     """Aiven Client with high-level operations"""
 
-    def get_service_versions(self) -> Mapping:
+    def get_service_versions(self) -> Sequence[Mapping[str, str]]:
         return self.verify(self.get, "/service_versions", result_key="service_versions")
 
-    def get_service_indexes(self, project: str, service: str) -> Mapping:
+    def get_service_indexes(self, project: str, service: str) -> Sequence:
```
dmitrii-vasilev, line 191: Can we be more specific with the result type here and in the following functions?

aiven/client/client.py
```diff
@@ -1647,10 +1647,10 @@ def update_account_authentication_method(
         account_id: str,
         authentication_id: str,
         method_name: Optional[str] = None,
-        method_enable: Optional[str] = None,
+        method_enable: Optional[bool] = None,
         options: Optional[Mapping[str, str]] = None,
     ) -> Mapping:
-        body = dict(options) if options else {}
+        body = dict(options) if options else {}  # type: Dict[str, Any]
```
dmitrii-vasilev, line 1653: Can we do `body: Dict[str, Any]` instead of the comment?

dmitrii-vasilev (dismissed):
Reviewed all the changes.

Wondering if we can be more specific about `Sequence` types in `client.py`?

aiven/client/cli.py
```diff
@@ -1391,7 +1409,7 @@ def _get_usage_from_args(self):
             usage = self.args.usage
         return usage
 
-    def _get_privatelink_connection_id_from_args(self):
+    def _get_privatelink_connection_id_from_args(self) -> Union[object, str]:
```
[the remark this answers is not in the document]
dangquangdon: `Literal` is only in 3.8+

aiven/client/cli.py
```diff
@@ -1533,7 +1551,7 @@ def service__cli(self):
                 raise
             raise argx.UserError("Executable '{}' is not available, cannot launch {} client".format(command, service_type))
 
-    def _build_influx_start_info(self, url):
+    def _build_influx_start_info(self, url: str) -> Tuple[str, List, Mapping]:
```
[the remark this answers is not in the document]
dangquangdon: I tried that at first, `mypy` doesn't like it. Because:
- the return value `params` will be later used in an concatenation expession in `self.service__cli()`, so it must be a `List`
- It can't be `List[str]` because it could contain `None` value. `info.hostname`, `info.port`, `info.username` could be `None`
- It can't be `List[Optional[str]]` because mypy doesn't allow `List[str] + List[Optional[str]]` :sweat: 

So I decided it keep it generic like this

aiven/client/cli.py
```diff
@@ -2659,12 +2677,12 @@ def service__task_get(self):
...
         default="delete",
     )
-    def service__topic_create(self):
+    def service__topic_create(self) -> None:
         """Create a Kafka topic"""
 
         tags = list(map(parse_tag_str, self.args.topic_option_tag or []))
         tag_keys = list(map(lambda d: d.get("key"), tags))
-        repeated_keys = [key for key, count in Counter(tag_keys).items() if count > 1]
+        repeated_keys = [key for key, count in Counter(tag_keys).items() if count > 1 and key]
```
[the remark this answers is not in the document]
dangquangdon: Because `Counter(tag_keys).items()` returns `key` as `str | None`. But perhaps it should be `key is not None` to not exclude the empty string.

aiven/client/cli.py
```diff
@@ -2787,12 +2806,12 @@ def service__acl_list(self):
...
     @arg("--username", help="Only show rules for user", required=False)
     @arg.json
-    def service__es_acl_list(self):
+    def service__es_acl_list(self) -> None:
         """List Elasticsearch ACL configuration"""
         response = self.client.list_service_elasticsearch_acl_config(
             project=self.get_project(), service=self.args.service_name
         )
-        acl_config = response.get("elasticsearch_acl_config")
+        acl_config = response.get("elasticsearch_acl_config", {})
```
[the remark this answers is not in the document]
dangquangdon: With `acl_config = response.get("elasticsearch_acl_config")`, `acl_config` could be `None | Any`. So mypy won't allow it to be passed in `self.print_response`. And also, because it's `None | Any`, it cannot use `.get()` in the later part of the function.
And also, the `elasticsearch_acl_config` from the response will be a dictionary, so I think use default `{}` is reasonable.

aiven/client/cli.py
```diff
@@ -3424,12 +3443,12 @@ def service__flink__job__cancel(self):
     @arg.project
     @arg("service", nargs="+", help="Service to wait for")
     @arg.timeout
-    def service__wait(self):  # pylint: disable=inconsistent-return-statements
+    def service__wait(self) -> Optional[int]:  # pylint: disable=inconsistent-return-statements
```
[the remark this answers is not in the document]
dangquangdon: `Literal` is only available in py3.8+

aiven/client/cli.py
```diff
@@ -3603,18 +3622,20 @@ def vpc__peering_connection__list(self):
...
         try:
             try:
-                peering_connection = self.client.get_project_vpc_peering_connection(
-                    project=project_name,
-                    project_vpc_id=self.args.project_vpc_id,
-                    peer_cloud_account=self.args.peer_cloud_account,
-                    peer_resource_group=self.args.peer_resource_group,
-                    peer_vpc=self.args.peer_vpc,
-                )
+                peering_connection = {
```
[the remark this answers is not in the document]
dangquangdon: I changed the return type from `client` as `Dict`

aiven/client/cli.py
dangquangdon: `Literal` is only in py3.8+

aiven/client/cli.py
dangquangdon: I'm using `dict()` instead, because `Mapping` doesn't have `copy()`

aiven/client/cli.py
dangquangdon: I'm using `dict()` instead, because Mapping doesn't have `copy()`

aiven/client/cli.py
dangquangdon: actually, depends on what `mode` the file is opened in. Here it is default to read, but it could be used with other mode too, so the return type is not necessarily just `TextIO`. I thought we could just use `IO` to be more generic

aiven/client/cli.py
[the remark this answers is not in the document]
dmitrii-vasilev: Ah, OK. Make sense. :+1:

aiven/client/cli.py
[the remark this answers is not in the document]
dmitrii-vasilev: Hmm, maybe set type for `tag_keys`? They can't be `None` nor empty

aiven/client/cli.py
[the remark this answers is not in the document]
dmitrii-vasilev: I'd convert it to `response["elasticsearch_acl_config"]` instead. Previously with `None` it would fail later on `acl_config.get()`. And with `{}` it will fail on `for acl in acl_config.get("acls")`. Both errors will be really obscure. I think it should explicitly fail if `acl_config` isn't found.

dmitrii-vasilev (approved):
LGTM

PR #288, opened 2022-05-09
Add further support for managing OAuth2 client apps configurations

ngilles-aiven:


m-novikov (dismissed):
Looks goot, but has issue with naming though. We generally use singluar forms for commands e.g. `user-list` so 
* `secrets-list` -> `secret-list`
* `redirects-list` -> `redirect-list`
Also verbs `create` instead of `add` and `delete` instead of `remove`

lornajane:
If you're adding commands to the client, could you also add them to the docs, at least as a reference even if we build the proper usage example later? https://developer.aiven.io/docs/tools/cli.html

Issue #290, opened 2022-05-30
Change default to `--no-project-vpc` when creating a service


TibsAtWork:
# What is currently missing?

When I create a service using the console, it defaults to "Public Internet".

When I create a service using `avn`, it defaults to using a VPC is there is one available. For instance I observed this with:
```
avn project switch <project>
avn service create tibs-demo-pg --service-type pg --cloud google-europe-north1 --plan hobbyist
```

When I then try to connect to the service CLI:
```
avn service cli tibs-demo-pg
```
I get:
```
psql: error: could not translate host name "tibs-demo-pg-dev-sandbox.aivencloud.com" to address: nodename nor servname provided, or not known
```
(and the same result if I use `psql` directly).

This was entirely surprising to me, as I'd expect the console and CLI to behave the same way when creating a service.

I can work around this by knowing what the problem is, and then migrating the service.
For reference of anyone reading this, that means:

1. Go to the *Cloud and VPC* section on the service page on the console
2. Check that it says "Project VPC"
3. Select *Migrate Cloud*
4. Choose a service from the appropriate tab (for instance, `Europe`) instead of the `VPC` tab
5. Migrate

# How could this be improved?

Change the default behaviour to pass `project_vpc_id: null`, as if the user had specified `--no-project-vpc`.

This is an incompatible change, so would need to be called out in the release notes.

# Is this a feature you would work on yourself?

[x] I plan to open a pull request for this feature

rikonen:
I wonder if a safer change would be to make service creation check whether the project has VPCs, and if it does, fail the operation unless either `--no-project-vpc` or `--project-vpc-id <vpc_id>` switch is given? Defaulting to either mode of operation is potentially unsafe / unexpected when there are VPCs in the project.

TibsAtWork:
I can see that point. I assume creating/deleting a VPC is a rare occurrence? (otherwise I'd worry about the race condition between asking if there's a VPC and doing the creation/update)

I see from `avn vpc list` that quite a few clouds have VPCs, so its quite possible for a user to come across this.

It does feel a little wrong to have to do a query before a create/update, but that's perhaps not an issue.

It is a bit horrible to tell the user "I shan't do that because of this thing you've never heard of, try again with `--no-project-vpc` or `--project-vpc-id <vpc_id>`" - the error message would need some work to make it friendly.

rikonen:
[...]

PR #291, opened 2022-05-30
Do not use a VPC by default

TibsAtWork:
Make `--no-project-vpc` the default, for consistency with service creation (and update) in the console.

Add a new flag `--use-project-vpc` to get back to the previous behaviour.

Fixes: https://github.com/aiven/aiven-client/issues/290

When a new service is created at the console, it will be created in "Public Internet" even if there is a VPC available.

When `avn service create` is used to create a new service, it will default to using a VPC if there is one available in the target cloud. This is unexpected, and can lead to (for instance):
```
avn service cli <pg-service-name>
```
giving an error like
```
psql: error: could not translate host name "<pg-service-name>" to address: nodename nor servname provided, or not known
```

This change alters the default to act like the console, by making `--no-project-vpc` the default. It also introduces a new switch to allow getting the original behaviour.

(It effectively makes the value `no_project_vpc` into a 3-state value. We want to be able to tell when the user explicitly specified either `--no-project-vpc` or `--use-project-vpc` so that the code can still give an error if they also specify `--project-vpc-id=<N>`.)

I'm not 100% happy about adding a new flag, but I assume we do want to be able to use the previous default behaviour, and I did not want to remove/rename the existing flag.

# Testing

There are new tests at the end of `tests/test_cli.py`. As this is an interaction between three switches,  it seemed important to test that they did work the way I expected.

I've also tested manually. Using project `dev-sandbox` I did:

* `python -m aiven.client service create --service-type pg --cloud google-europe-north1 --plan hobbyist tibs-pg-vanilla`

   This created a service with "Public Internet", which is the desired result of this PR.

* `python -m aiven.client service create --service-type pg --cloud google-europe-north1 --plan hobbyist tibs-pg-novpc --no-project-vpc`

  This also created a service with "Public Internet", obeying the existing switch.

* `python -m aiven.client service create --service-type pg --cloud google-europe-north1 --plan hobbyist tibs-pg-vpc --use-project-vpc`

  This created a service with "Project VPC", obeying the new switch.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -3786,7 +3786,8 @@ ssl.truststore.type=JKS
         action="store_const",
         const=True,
         default=UNDEFINED,
-        help="Do not put the service into a project VPC even if the project has one in the selected cloud. This is the default action",
+        help="Do not put the service into a project VPC even if the project has one in the selected cloud."
+        " This is the default action",
     )
     @arg(
         "--use-project-vpc",
@@ -4051,7 +4052,8 @@ ssl.truststore.type=JKS
         action="store_const",
         const=True,
         default=UNDEFINED,
-        help="Do not put the service into a project VPC even if the project has one in the selected cloud. This is the default action",
+        help="Do not put the service into a project VPC even if the project has one in the selected cloud."
+        " This is the default action",
     )
     @arg(
         "--use-project-vpc",
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -3778,8 +3778,17 @@ ssl.truststore.type=JKS
     )
     @arg(
         "--no-project-vpc",
-        action="store_true",
-        help="Do not put the service into a project VPC even if the project has one in the selected cloud",
+        action="store_const",
+        const=True,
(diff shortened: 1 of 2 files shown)


rikonen:
As noted in https://github.com/aiven/aiven-client/issues/290#issuecomment-1141901380, this mode of operation can be problematic as well because it can be equally surprising for people who have intentionally created the VPC and then try creating a service there for the service not to appear in the VPC, especially as that was the old behavior. Prompting the user to explicitly define the expected mode of operation when VPCs exist would ensure nobody gets surprised by the results.

TibsAtWork:
I think the point made by @rikonen is an important one. They also point out on the issue that there is a potential security surprise if someone expects the old behaviour (automatic use of the VPC) and gets the new behaviour (public internet).

I'll investigate the proposed alternative solution, and may end up closing this PR if it seems a better approach.

TibsAtWork:
Closing because I think the approach suggested by @rikonen is important, and the appropriate changes are also simpler. See https://github.com/aiven/aiven-client/pull/292 for the replacement PR.

PR #292, opened 2022-05-31
Ask user to be specific when there is a VPC

TibsAtWork:
If the user is creating/updating a service in a cloud that has a VPC, and they have specified the cloud with `--cloud`, and they have not specified either `--project-vpc-id=<id>` or `--no-project-vpc`, then stop with an error asking them to do one or the other.

Fixes: https://github.com/aiven/aiven-client/issues/290

When a new service is created at the console, it will be created in "Public Internet" even if there is a VPC available.

When `avn service create` is used to create a new service, it will default to using a VPC if there is one available in the target cloud. This is unexpected, and can lead to (for instance):
```
avn service cli <pg-service-name>
```
giving an error like
```
psql: error: could not translate host name "<pg-service-name>" to address: nodename nor servname provided, or not known
```

This change refuses to do either when there is a VPC for the requested cloud - the user is requested to be specific about which alternative they wanted.

If the user doesn't specify a cloud (which is common for `service update` when the service is not being moved, and allowed for service creation), then we don't require the VPC switches.

**Note:** There are some additional changes to make `mypy` happy with the existing code - I've added `-> None` in appropriate places.

# Testing

There are new tests at the end of `tests/test_cli.py`. Given a change of this nature, actual tests seemed necessary. In the longer term, it may not be necessary to keep them, as this functionality is unlikely to change.

I've also tested manually. Using project `dev-sandbox` I did:

* `python -m aiven.client service create --service-type pg --cloud google-europe-north1 --plan hobbyist tibs-pg-vanilla`

  That says
  ```
  ERROR	command failed: UserError: Cloud google-europe-north1 has a VPC. Specify --project-vpc-id=<vpc-id> to use it, or --no-project-vpc to use the public internet
  ```
  which is what we want.
 
* `python -m aiven.client service create --service-type pg --cloud google-europe-north1 --plan hobbyist tibs-pg-novpc --no-project-vpc`

  That created a service with "Public Internet", obeying the switch.

* `python -m aiven.client service create --service-type pg --cloud google-europe-north1 --plan hobbyist tibs-pg-vpc --project-vpc-id=c6a08e40-778c-40b3-bfaf-17a3d3800ab7`

  That created a service with "Project VPC", obeying the switch

* `python -m aiven.client service create --service-type pg --cloud google-europe-west2 --plan hobbyist tibs-pg-also-no-vpc`

  That created a service with "Public Internet", as `google-europe-west2` is not in the list reported by `avn vpc list`

Further testing:

* `python -m aiven.client service update tibs-pg-novpc --cloud google-europe-west1`

  This says:
  ```
  ERROR	command failed: UserError: Cloud google-europe-west1 has a VPC. Specify --project-vpc-id=<vpc-id> to use it, or --no-project-vpc to use the public internet
  ```
  which seems sensible as we don't know whether the user wanted a VPC on that new cloud or not.

* `python -m aiven.client service update tibs-pg-novpc --plan startup-4`

  This should just change the plan - it doesn't matter whether there is a VPC in the cloud if we're not changing that. And it behaves as expected.

* `python -m aiven.client service create --service-type pg --plan hobbyist tibs-pg-guess`

  When we create a new service without specifying the cloud, we expect to get some sort of cached decision for the cloud. In this case, it created a service in `europe-north1` with "Project VPC". Since it's not documented what happens in this case, I'm happy to ignore it for the purpose of this PR. (The previous `create` was in `google-europe-west2`, and the one before that was `google-europe-north1`, with VPC, so this what to expect is not obvious, and I assume is related to how the backend service works.)

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -3707,9 +3707,31 @@ ssl.truststore.type=JKS
             print(ex.response.text)
             raise
 
+    def _validate_using_cloud_vpc(self) -> None:
+        """Raise an error if we might unexpectedly end up using a Project VPC"""
+        # If the user was specific about VPC then we don't need to worry
+        if self.args.no_project_vpc or self.args.project_vpc_id is not None:
+            return
+
+        # If the user did not specify a cloud, then `service create` will use
+        # some variant of something historical, and `service update` will use
+        # the same cloud. In either case, we don't need to worry.
+        if self.args.cloud is None:
+            return
+
+        project_name = self.get_project()
+        vpc_list = self.client.list_project_vpcs(project=project_name)["vpcs"]
+        clouds_with_vpc = [x["cloud_name"] for x in vpc_list]
+        if self.args.cloud in clouds_with_vpc:
+            raise argx.UserError(
+                f"Cloud {self.args.cloud} has a VPC. Specify --project-vpc-id=<vpc-id> to use it,"
+                " or --no-project-vpc to use the public internet"
+            )
+
     def _get_service_project_vpc_id(self) -> Optional[Union[object, str]]:
         """Utility method for service_create and service_update"""
         if self.args.project_vpc_id is None:
+            self._validate_using_cloud_vpc()
             project_vpc_id = None if self.args.no_project_vpc else UNDEFINED
         elif self.args.no_project_vpc:
             raise argx.UserError("Only one of --project-vpc-id and --no-project-vpc can be specified")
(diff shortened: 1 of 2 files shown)


rikonen (changes requested):
Looks good to me for the service creation part but for service update this would now cause an error when not explicitly specifying the VPC related parameters even if also cloud is not specified. That would be annoying as the old behavior in that case would've been not to touch the VPC/cloud config at all and only update whatever fields were provided but now you'd be forced to always pass the extra parameters.

TibsAtWork:
@rikonen I'm back from holiday, and also back from being ill after that (not related!), so would appreciate you and the Aiven Client team looking at this to see if it makes sense.

rikonen:
Latest version looks good logic-wise but should remove the debug prints and squash commits.

rikonen (approved):

PR #296, opened 2022-07-25
Make package PEP 561 compatible

rominf:

Even though package was fully type annotated, it was not PEP 561 compatible:
https://mypy.readthedocs.io/en/latest/installed_packages.html#creating-pep-561-compatible-packages

This change fixes the issue.


dmitrii-vasilev (approved):
LGTM

PR #297, opened 2022-08-11
Adding connection-info kcat support in aiven-client

ftisiot:

Kafkacat is the old name of the tool now named [kcat](https://github.com/edenhill/kcat), the PR adds the functions to retrieve the kcat call, we'll be then able to deprecate the `kafkacat` one and remove it safely some time in the future

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -1343,6 +1343,7 @@ class AivenCLI(argx.CommandLineTool):
             store = Store.skip
         return store
 
+    # This method should be deprecated and removed over time. Check service__connection_info__kcat
     @arg.project
     @arg.service_name
     @arg("-r", "--route", choices=("dynamic", "privatelink", "public"))
@@ -1366,7 +1367,8 @@ class AivenCLI(argx.CommandLineTool):
                 privatelink_connection_id=self._get_privatelink_connection_id_from_args(),
                 username=self.args.username,
             )
-            cmd = ci.kafkacat(
+            cmd = ci.kcat(
+                "kafkacat",
                 store,
                 get_project_ca=self._get_project_ca,
                 ca_path=self.args.ca_path,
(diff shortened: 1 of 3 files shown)


rominf:
Converted to draft since tests do not pass to prevent accidental merge.

ftisiot:
Hi! In order to test this please find a Aiven for Kafka service and execute

```
avn service connection-info kcat  [SERVICE_NAME]
```

The output should be similar to 

```
kcat -b HOST:PORT -X security.protocol=SSL -X ssl.ca.location=ca.pem -X ssl.key.location=service.key -X ssl.certificate.location=service.crt
```

If the Aiven for Apache Kafka uses SASL authentication you can use the `-a sasl` flag and you should get something like

```
kcat -b HOST:PORT -X security.protocol=SASL_SSL -X ssl.ca.location=ca.pem -X sasl.mechanisms=SCRAM-SHA-256 -X sasl.username=avnadmin -X sasl.password=PASSWORD
```

rominf (changes requested):
Please add user-friendly message in case Kafka has SASL disabled.

Now it's not user-friendly:
```
avn service connection-info kcat -a sasl [service-without-sasl]
ERROR   command failed: ConnectionInfoError: Could not find connection information with filters route=dynamic, privatelink_connection_id=<object object at 0x7f2a939b4c20>, kafka_authentication_method=sasl
```

aiven/client/cli.py
```diff
@@ -1343,6 +1343,7 @@ def _get_store_from_args(self) -> Store:
             store = Store.skip
         return store
 
+    # This method should be deprecated and removed over time. Check service__connection_info__kcat
```
HelenMel, line 1346: I would rather use `warnings.warn("Use the kcat method instead", DeprecationWarning)`

HelenMel:
Good PR. I have one minor comment

aiven/client/cli.py
ftisiot: thanks!

aiven/client/connection_info/kafka.py
```diff
@@ -88,19 +104,25 @@ def from_service(
...
+                route=route,
+                privatelink_connection_id=privatelink_connection_id,
+                kafka_authentication_method="sasl",
+            )
+            
+            user = find_user(service, username)
+            if "password" not in user:
+                raise ConnectionInfoError(f"Could not find password for username {username}")
+            return cls(host=info["host"], port=info["port"], username=username, password=user["password"])
+        except ConnectionInfoError:
```
rominf, line 118: If the password is missing, then you'll get `SASL auth...` error, which is wrong. I think the better way is to check user config via separate call to understand if SASL auth is enabled.

rominf (changes requested):

aiven/client/connection_info/kafka.py
ftisiot: gotcha, doing for both SSL and SASL methods

aiven/client/connection_info/kafka.py
```diff
@@ -52,7 +68,15 @@ def from_service(
             raise ConnectionInfoError(
                 "Cannot format kafka connection info for service type {service_type}".format_map(service)
             )
-
+        try:
+            find_component(
+                service["components"],
+                kafka_authentication_method="certificate",
+            )
```
rominf, line 75: You have two `find_component` calls with different arguments with the same `kafka_authentication_method="certificate"`. Why not keep just one?

ftisiot: I believe a standalone `find_component` checking only the `kafka_authentication_method` checks if the authentication method is enabled, while the `find_component` checking both `kafka_authentication_method` and `route` checks that the authentication method is enabled in the route specified?

rominf: Ah, yeah, you are right. Thanks.

rominf (approved):
LTGM. I'll let @HelenMel check this before merging.

ftisiot:
@HelenMel could you please double check that everything is ok from your side?

HelenMel (approved):

PR #298, opened 2022-08-17
Temporary pin pip version in GitHub Actions to allow our way of versioning

rominf:

pip==21.2.4 is the latest version that doesn't enforce PEP 440.

This is a temporary solution to unblock PRs.


rominf:
This is a prerequisite for #297.

dmitrii-vasilev (approved):
LGTM

PR #300, opened 2022-08-17
Add Fedora 37 to GitHub Actions

rominf:

We use Fedora 37 downstream, so to find bugs earlier, let's test Aiven Client on Fedora 37 too.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -75,3 +75,33 @@ jobs:
           pip install -e .
           pytest -vv tests/
 
+  test-fedora:
+    runs-on: ubuntu-latest
+    container: 'fedora:37'
+    needs: lint
+    steps:
+      - id: dependencies
+        run: sudo dnf install -y git-core make
+
+      - id: checkout-code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -75,3 +75,33 @@ jobs:
...
+        name: Build and test
+        run: |
+          make build-dep-fedora
+          make rpm
+
+      - id: install
+        name: Install
+        run: sudo dnf install -y rpms/noarch/*.rpm
+
+      # test step is not needed (tests run on a build)
```
dmitrii-vasilev, line 107: Hmm, on build we run `make build-dep-fedora rpm`. I don't think we run tests there

rominf: `make rpm` executes pytest. We can change it, but it should be a separate PR.

dmitrii-vasilev: I don't think it does. Also, `make rpm` takes about 2 min on my machine. Really not enough to run the tests. :laughing:

rominf: Also (https://github.com/aiven/aiven-client/actions/runs/4014851129/jobs/6895903606):
image

dmitrii-vasilev: Yes, you're right. I was looking in a wrong place.

dmitrii-vasilev (approved):
LGTM

PR #303, opened 2022-09-01
Enable and disable authorization in Kafka REST proxy

juha-aiven:
Add support for enabling and disable Kafka REST proxy authorization in a Kafka service. Adds two switches `--disable-rest-proxy-authorization` and `--enable-rest-proxy-authorization` to `service update` command.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -4049,6 +4049,30 @@ ssl.truststore.type=JKS
                 )
             ) from ex
 
+    def _schema_registry_authorization(self) -> Optional[bool]:
+        if self.args.enable_schema_registry_authorization and self.args.disable_schema_registry_authorization:
+            raise argx.UserError(
+                "--enable-schema-registry-authorization and --disable-schema-registry-authorization are mutually exclusive."
+            )
+        schema_registry_authorization = None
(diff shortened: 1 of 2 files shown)


juha-aiven:
The build is failing because of this

```
Run make pylint
python3 -m pylint aiven tests
************* Module aiven.client.cli
aiven/client/cli.py:4814:19: W3101: Missing timeout argument for method 'requests.post' can cause your program to hang indefinitely (missing-timeout)
```

The PR doesn't touch line 4814 of `aiven/client/cli.py`.

On my laptop `make pylint` works both in `main` and in this branch.

tvainika (changes requested):
This is wrong approach.

There is service update command with `-c` flag to apply configuration updates. The authorization should be enabled by running `avn service update -c somekey=true` instead of new additional long options to enable/disable this specific configuration. The generic configuration feature should be used instead of polliting argument space.

jjaakola-aiven:
@tvainika comment is correct. The `avn service update` is generic for all services.
IMO these switches should be removed `--use-karapace`, `--enable-schema-registry-authorization`, `--disable-schema-registry-authorization` and new service specific arguments should not be introduced.

PR #315, opened 2023-01-17
Remove Flink beta service API

jjaakola-aiven:

Existing Flink beta services will be discontinued and will be replaced with new Flink experience on Aiven.


juha-aiven (approved):
Looks good. I grepped `flink` in both modified files and no hit.

PR #316, opened 2023-01-18
client: add option to set ip_filter 'description' field via avn cli […

basemeraphiona:

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -13,6 +13,7 @@ from aiven.client.connection_info.redis import RedisConnectionInfo
 from aiven.client.pretty import TableLayout
 from aiven.client.speller import suggest
 from argparse import ArgumentParser
+from ast import literal_eval
 from collections import Counter
 from datetime import datetime, timedelta, timezone
 from decimal import Decimal
@@ -42,14 +43,14 @@ AUTHENTICATION_METHOD_COLUMNS = [
 EOL_ADVANCE_WARNING_TIME = timedelta(weeks=26)  # Give 6 months advance notice for EOL services
 
 
-def convert_str_to_value(schema: Mapping[str, Any], str_value: Optional[str]) -> Any:
-    if str_value is not None:
-        if "string" in schema["type"]:
-            return str_value
+def convert_str_to_value(schema: Mapping[str, Any], value: Optional[Any]) -> Any:
+    if value is not None:
+        if "string" in schema["type"] or "object" in schema["type"]:
+            return value
         elif "integer" in schema["type"]:
-            return int(str_value, 0)  # automatically convert from '123', '0x123', '0o644', etc.
+            return int(value, 0)  # automatically convert from '123', '0x123', '0o644', etc.
         elif "number" in schema["type"]:
-            return float(str_value)
+            return float(value)
         elif "boolean" in schema["type"]:
             values = {
                 "1": True,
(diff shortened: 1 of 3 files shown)


aiven/client/cli.py
```diff
@@ -64,7 +64,13 @@ def convert_str_to_value(schema: Mapping[str, Any], str_value: Optional[str]) ->
                     "Invalid boolean value {!r}: expected one of {}".format(str_value, ", ".join(values))
                 ) from ex
         elif "array" in schema["type"]:
-            return [convert_str_to_value(schema["items"], val) for val in str_value.split(",")]
+            str_value_array = []
```
heikju, line 67: This seems unnecessary, we can do `str_value_array = [jsonlib.loads(str_value)]` on line 70 for the same end result.

tests/test_cli.py
```diff
@@ -140,6 +140,86 @@ def test_create_user_config() -> None:
     }
 
 
+@pytest.mark.parametrize(
+    ("user_config_args,config_type,str_value"),
+    [
+        ('{"description":"test","network":"0.0.0.0/0"}', "array", [{"description": "test", "network": "0.0.0.0/0"}]),
```
heikju, line 146: If a single JSON object as a string is parsed into a one element array, then what kind of input is needed for a multi-element one (i.e. multiple IP filters with descriptions)? It's not immediately obvious to me when reading the above code, so I'd like to at least have a test case for that scenario.

tests/test_cli.py
```diff
@@ -140,6 +140,86 @@ def test_create_user_config() -> None:
...
+@pytest.mark.parametrize(
+    ("user_config_args,cnfig_type,error_message"),
+    [
+        ("0.0.0.0/8,0.0.0.0", "dict", "Support for option value type(s) 'dict' not implemented"),
+        ("True", "boolean", "Invalid boolean value 'True': expected one of 1, 0, true, false"),
+    ],
+)
+def test_convert_str_to_value_fails(user_config_args: str, cnfig_type: str, error_message: str) -> None:
+    schema = {
+        "type": cnfig_type,
```
heikju, line 194: This is `config_type` in the other test. Typo?

tests/test_cli.py
```diff
@@ -140,6 +140,86 @@ def test_create_user_config() -> None:
...
+    ("user_config_args,config_type,str_value"),
+    [
+        ('{"description":"test","network":"0.0.0.0/0"}', "array", [{"description": "test", "network": "0.0.0.0/0"}]),
+        ("0.0.0.0,0.0.0/8", "array", ["0.0.0.0", "0.0.0/8"]),
+        ("123", "integer", 123),
+        ("10.0", "number", 10.0),
+        ("false", "boolean", False),
+    ],
+)
+def test_convert_str_to_value(user_config_args: str, config_type: str, str_value: Union[str, dict]) -> None:
```
heikju, line 153: I think `str_value` is not a great name for the expected value when it can be of other types too, and it seems it can also be an array contrary to what the type hint states. I personally would be happy with `expected_value: Any` in a test context, but if you feel like being more pedantic then that's fine too.

heikju (changes requested):

aiven/client/cli.py
basemeraphiona: yes but if we break it up and have a `return` on line 70 and line 72 `mypy` complains about having too many `return` statements. Unless there is another way of approaching it without having two `return` statements.

tests/test_cli.py
basemeraphiona: It would fail. I think we can have a test case for that

heikju:
Thanks, the new tests are good to have and make clear enough how that particular setting works,
```
python3 -m aiven.client.cli service update -c 'ip_filter=[{"network": "0.0.0.0/0", "description": "Allow all"}]' <service>
```
worked for me.

aiven/client/cli.py
```diff
@@ -44,7 +45,7 @@
 
 def convert_str_to_value(schema: Mapping[str, Any], str_value: Optional[str]) -> Any:
     if str_value is not None:
-        if "string" in schema["type"]:
+        if "string" in schema["type"] or "object" in schema["type"]:
```
heikju, line 48: Though this seems a bit weird, is it now so that `str_value` can also be of object type? If so, I'd like to remove the `str_` prefix from name and have an accurate type hint for the arg.

heikju (approved):

PR #320, opened 2023-01-20
cli:add option to create fork from PITR to Aiven CLI

basemeraphiona:
Add option to define the PITR recovery point when creating a fork via Aiven cli


aiven/client/cli.py
```diff
@@ -4037,6 +4039,13 @@ def service__create(self) -> None:
                         "source_service": self.args.read_replica_for,
                     }
                 )
+        elif self.args.recovery_target_time and self.args.source_service_name:
+            if self.args.service_type in ("pg", "myqsl"):
```
heikju, line 4043: If `recovery_target_time` is valid only for some service types, then I think `service create` command should fail if it's given for a service of some other type. This way, we don't create a service in a state other than what the user is expecting. Better for them to be informed that recovery to a specific time is not possible for the forked service.

heikju (dismissed):

aiven/client/cli.py
basemeraphiona: I am thinking maybe we let this fail at the `api` so that incase we add support for PITR for another service besides `pg` and `mysql` then we don't have to come and make a change here. Opinion on this?

heikju: Yes, that would be my preference.

aiven/client/cli.py
```diff
@@ -3997,6 +3997,8 @@ def vpc__user_peer_network_cidr__delete(self) -> None:
         default=False,
         help="Enable termination protection",
     )
+    @arg("--source-service-name", help="Service to fork from")
+    @arg("--recovery-target-time", help="PITR recovery point")
```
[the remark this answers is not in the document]
heikju: And also, I'd like to know if it's really necessary to name the command line arg differently from the user_config setting, especially since the help text here uses the exact same words as the setting name. Would `--service-to-fork-from` be wrong?

aiven/client/cli.py
```diff
@@ -1779,14 +1779,14 @@ def service__connection_pool_create(self) -> None:
...
-            pool_mode=self.args.pool_mode,
-        )
+        kwargs = {
+            "project": self.get_project(),
+            "service": self.args.service_name,
+            "pool_name": self.args.pool_name,
+            "dbname": self.args.dbname,
+            "pool_size": self.args.pool_size,
+            "pool_mode": self.args.pool_mode,
+        }
```
dmitrii-vasilev, line 1789: Hmm, why this change? It looks OK and arguable more readable with `dict()`

aiven/client/cli.py
```diff
@@ -3863,6 +3865,11 @@ def service__create(self) -> None:
                         "source_service": self.args.read_replica_for,
                     }
                 )
+        elif self.args.recovery_target_time and self.args.service_to_fork_from:
+            user_config["service_to_fork_from"] = self.args.service_to_fork_from
+            user_config["recovery_target_time"] = self.args.recovery_target_time
+        elif self.args.service_to_fork_from:
+            user_config["service_to_fork_from"] = self.args.service_to_fork_from
```
dmitrii-vasilev, line 3872:
```suggestion
        elif self.args.service_to_fork_from:
            user_config["service_to_fork_from"] = self.args.service_to_fork_from
            if self.args.recovery_target_time:
                user_config["recovery_target_time"] = self.args.recovery_target_time
```

dmitrii-vasilev: Also I'm wondering where we validate `user_config`?

aiven/client/cli.py
```diff
@@ -3823,6 +3823,8 @@ def vpc__user_peer_network_cidr__delete(self) -> None:
         default=False,
         help="Enable termination protection",
     )
+    @arg("--service-to-fork-from", help="Service to fork from")
+    @arg("--recovery-target-time", help="PITR recovery point in format 2023-02-01 11:38:49")
```
dmitrii-vasilev, line 3827:
```suggestion
    @arg("--recovery-target-time", help="PITR recovery point datetime in ISO 3166 format")
```

dmitrii-vasilev (changes requested):
Some comments

aiven/client/cli.py
heikju: I think that's the job for the backend. That makes it less likely for us to need to update this client if we add/change somehting there.

I personally would just put all values given on command line to the dict unconditionally. If some combination of them is invalid, then it's the APIs job to tell us.

aiven/client/cli.py
heikju: I disagree with this. A concrete example tells me immediately what to do, no need to force the user to study standards.

aiven/client/cli.py
basemeraphiona: this change was because of mypy it was complaining about the previous approach

dmitrii-vasilev: Hmm, wondering what was the complain?

dmitrii-vasilev (approved):
LGTM

aiven/client/cli.py
basemeraphiona: This was fixed last week so I don't need to make the change here anymore

PR #321, opened 2023-01-24
Add new Flink developer experience

jjaakola-aiven:

Aiven Flink has new developer experience. This adds the API to aiven-client.


aiven/client/client.py
```diff
@@ -2134,3 +2135,410 @@ def delete_oauth2_client_secret(self, account_id: str, client_id: str, secret_id
...
+                "application",
+            ),
+        )
+
+    def flink_create_application(
+        self,
+        *,
+        project: str,
+        service: str,
+        application_properties: Mapping[str, str],
```
juha-aiven, line 2162: The type is not `Mapping[str, str]`. If the key is `application_version`, the value is an embedded object.

https://api.aiven.io/doc/#tag/Service:_Flink/operation/ServiceFlinkCreateApplication

aiven/client/client.py
```diff
@@ -2134,3 +2135,410 @@ def delete_oauth2_client_secret(self, account_id: str, client_id: str, secret_id
             self.delete,
             self.build_path("account", account_id, "oauth_client", client_id, "secret", secret_id),
         )
+
+    def flink_list_applications(
+        self,
+        *,
+        project: str,
+        service: str,
+    ) -> Mapping:
```
juha-aiven, line 2144: Add key and value types to`Mapping`.

aiven/client/client.py
```diff
@@ -2134,3 +2135,410 @@ def delete_oauth2_client_secret(self, account_id: str, client_id: str, secret_id
...
+        self,
+        *,
+        project: str,
+        service: str,
+        application_id: str,
+    ) -> Mapping:
+        try:
+            application_uuid = UUID(application_id)
+        except ValueError as e:
+            raise ValueError("Application id must be a valid UUID") from e
```
juha-aiven, line 2187: Better would be to have the exact param name in the error, e.g.`application_id not a valid UUID`

Is it needed to parse and format as string the `application_id`? The REST API will respond with an error if `application_id` is not an uuid? There's no validation on the client side of other parameters, e.g. `project` has certain naming convention, but it's not checked here - why check `application_id`?

juha-aiven (dismissed):
Typing of `Mapping` needs fixing. Maybe the `UUID` parsing should be just removed.

aiven/client/client.py
jjaakola-aiven: Done, I changed to `str, Any`.

aiven/client/client.py
jjaakola-aiven: Done, used `Any, Any` for all.

aiven/client/client.py
jjaakola-aiven: For consistency, removed the validation from client.

aiven/client/cliarg.py
```diff
@@ -204,3 +204,7 @@ def wrapped(self: CommandLineTool) -> T:
 
 arg.billing_email = arg("--billing-email", action="append", help="Billing email address")
 arg.tech_email = arg("--tech-email", action="append", help="Tech email address")
+
+arg.flink_application_id = arg("--application-id", required=True, help="Flink application id")
+arg.flink_application_version_id = arg("--application-version-id", required=True, help="Flink application id")
```
juha-aiven, line 209: `Flink application version id`

aiven/client/cli.py
```diff
@@ -5262,6 +5262,219 @@ def account__oauth2_client__secret_delete(self) -> None:
...
+                deployment_id=self.args.deployment_id,
+            ),
+        )
+
+    @arg.project
+    @arg.service_name
+    @arg.flink_application_id
+    @arg.flink_deployment_id
+    def service__flink__stop_application_deployment(self) -> None:
+        """Delete Flink application deployment"""
```
juha-aiven, line 5453: `Stop Flink application deployment`

juha-aiven (changes requested):
A few typos

aiven/client/cliarg.py
jjaakola-aiven: Done

aiven/client/cli.py
jjaakola-aiven: Done

juha-aiven (approved):
LGTM

PR #324, opened 2023-01-26
Build against python 3.11

rominf:

Build against python 3.11.


dmitrii-vasilev (approved):
LGTM

PR #325, opened 2023-01-26
switch to pyproject.toml

shinespb:

Switch to modern way of Python packaging.


rominf (changes requested):
I have some comments, I'll write them down after you prove that it works downstream. Marking as "Changes requested" to prevent merge.

Makefile
```diff
@@ -1,14 +1,12 @@
 short_ver = 2.18.0
 long_ver = $(shell git describe --long 2>/dev/null || echo $(short_ver)-0-unknown-g`git describe --always`)
```
rominf, line 2: No longer needed.

Makefile
```diff
@@ -57,13 +55,14 @@ build-dep-fedora:
...
 		python3-types-requests \
+		python3-setuptools_scm \
 		rpmdevtools \
 		tar
 
 .PHONY: rpm
 rpm: $(RPM)
 
-$(RPM): $(generated)
+$(RPM): aiven/client/version.py
```
rominf, line 65: Everything related to RPM should be dropped in my opinion. It's not the correct way of building RPMs. If we want to make RPMs for everybody, we should use [OBS](https://openbuildservice.org/) or something similar.

pyproject.toml
```diff
@@ -2,6 +2,65 @@
...
+readme = "README.rst"
+requires-python = ">=3.7"
+classifiers=[
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Topic :: Software Development :: Libraries",
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
```
rominf, line 24: Please add 3.10, 3.11, we have tests for them.

pyproject.toml
```diff
@@ -2,6 +2,65 @@
...
+    "mypy>=0.931,<1.0",
+    # Lock types-requests to the same major version used in downstream build environments
+    "types-requests>=2.26.1,<3.0",
+    # Lock pylint to the same major version used in downstream build environments
+    "pylint>=2.12.2,<3.0",
+    "pytest",
+    # Lock isort to the same major version used in downstream build environments
+    "isort>=5.10.1,<6.0",
+    # To prevent black from breaking (https://github.com/psf/black/issues/2964, https://github.com/psf/black/pull/2966)
+    "click < 8.1.0",
```
rominf, line 48: Please update versions to actual downstream versions.

rominf, line 48: This is not relevant anymore if you do the update (see the comment above).

pyproject.toml
```diff
@@ -2,6 +2,65 @@
...
+]
+
+
+[project.urls]
+"Homepage" = "https://github.com/aiven/aiven-client"
+"Bug Tracker" = "https://github.com/aiven/aiven-client/issues"
+
```
rominf, line 55: I would add a link to the docs here: https://docs.aiven.io/docs/tools/cli

rominf (changes requested):

Makefile
shinespb: does it maybe make sense also to remove `test-fedora` step from tests?

rominf: Why? We use Fedora downstream, so tests won't hurt.

rominf (approved):
LGTM. To be on the safe side, please update the downstream PR with Aiven Client update. I'll merge this PR after that.

ezotrank (approved):

aiven/client/cli.py
```diff
@@ -138,7 +138,6 @@ class ClientFactory(Protocol):  # pylint: disable=too-few-public-methods
         def __call__(self, base_url: str, show_http: bool, request_timeout: Optional[int]) -> client.AivenClient:
             ...
 
-
```
ezotrank, line 141: Should that change has been reverted?

rominf: I believe it's black version update which is guilty for this change. Otherwise linters won't pass.

aiven-client.spec
```diff
@@ -1,13 +1,24 @@
-Name:           aiven-client
+%global project aiven-client
+
+Name:           %{project}
```
paulwouters, line 3: note, probably this python package should end up being called python3-aiven-client to be fully fedora python guidelines compatible

pyproject.toml
```diff
@@ -2,6 +2,67 @@
...
+classifiers=[
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Topic :: Software Development :: Libraries",
+    "License :: OSI Approved :: Apache Software License",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
paulwouters, line 26: no python 3.12 ?

pyproject.toml
```diff
@@ -2,6 +2,67 @@
...
+    # Lock mypy to the same major version used in downstream build environments
+    "mypy>=0.991,<1.0",
+    # Lock types-requests to the same major version used in downstream build environments
+    "types-requests>=2.26.1,<3.0",
+    # Lock pylint to the same major version used in downstream build environments
+    "pylint>=2.15.5,<3.0",
+    "pytest",
+    # Lock isort to the same major version used in downstream build environments
+    "isort>=5.10.1,<6.0",
+    "click",
```
paulwouters, line 49: in general, locking to older versions for no good reason other than "prevent issues", is not recommended, It just moves the problems into the future.

paulwouters (approved):
LGTM, some comments left to consider in the future

PR #328, opened 2023-02-13
Add a command to change user password

heikju:

We have a pw change API, but client was missing a command for it. Add one.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -437,6 +437,14 @@ class AivenCLI(argx.CommandLineTool):
         message = self.client.expire_user_tokens()["message"]
         print(message)
 
+    @arg.verbose
+    def user__password_change(self) -> None:
+        """Change the password of the currently logged in user"""
+        current_password = getpass.getpass("Current password: ")
+        new_password = self.enter_password("New password: ", confirm=True)
+        self.client.change_user_password(current_password, new_password)
+        print("Password changed. Next, use 'avn user login' to log in again using your new password.")
+
     @arg("--description", required=True, help="Description of how the token will be used")
     @arg("--max-age-seconds", type=int, help="Maximum age of the token, if any")
     @arg(
--- a/aiven/client/client.py
+++ b/aiven/client/client.py
@@ -1762,6 +1762,13 @@ class AivenClient(AivenClientBase):
     def expire_user_tokens(self) -> Mapping:
         return self.verify(self.post, "/me/expire_tokens")
 
+    def change_user_password(self, current_password: str, new_password: str) -> Mapping:
+        request = {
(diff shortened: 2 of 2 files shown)


aiven/client/cli.py
```diff
@@ -437,6 +437,14 @@ def user__tokens_expire(self) -> None:
         message = self.client.expire_user_tokens()["message"]
         print(message)
 
+    @arg.verbose
+    def user__password_change(self) -> None:
+        """Change the password of the currently logged in user"""
+        current_password = getpass.getpass("Current password:")
+        new_password = self.enter_password("New password:", confirm=True)
```
rominf, line 444: Missing space after colon (in both lines). Also, I would add ` (hidden)` to make it clear that the password is not displayed at all.

aiven/client/cli.py
```diff
@@ -437,6 +437,14 @@ def user__tokens_expire(self) -> None:
...
+        message = self.client.change_user_password(current_password, new_password)
+        print(message)
```
rominf, line 446: In case of password reuse it print me:
```
ERROR   command failed: Error: {"errors":[{"message":"Password reuse is not allowed","status":403}],"message":"Password reuse is not allowed"}
```
Which is correct, but not user friendly. I believe:
```
ERROR   command failed: Password reuse is not allowed
```
is much more readable and enough for the user.

rominf: On success it prints the token. For me, it's unexpected behavior. Even more: when I try to do something, it shows me:
```
ERROR   command failed: Error: {"errors":[{"message":"Expired db token","status":401}],"message":"Expired db token"}
```
It would be much easier for the user if it showed: `Password has been changed` and then did login automatically.

rominf (changes requested):

aiven/client/cli.py
heikju: I think error logging happens here https://github.com/aiven/aiven-client/blob/main/aiven/client/argx.py#L334 for all commands. I'd rather not make it different for just one command, given that I don't agree 100% with the user-friendliness argument either: shorter non-JSON message may look less intimidating, but if customer reports a failure to do something, it contains less information to enable us to resolve the problem.

szypulka (dismissed):
Looks good to me, leave it unmerged to fix the mistype with space after colon

heikju:
Updated to add the spaces, and added a message instructing them to log in again since password change invalidates their current token. Requiring a new login after a password change is a common pattern I've seen, and it allows the user to immediately notice if they pw change was successful, or if they e.g. managed to mistype twice or copy&paste the wrong thing. I think this is friendly enough now.

rominf (approved):

PR #329, opened 2023-02-15
client: fix log level

alxric:

In the current implementation, if show_http is False, level ends up as NOTSET, which causes all messages to be processed.

By defaulting to INFO, we will hide all the debug messages unless show_http excplicitly is enabled.


rominf (approved):

szypulka (approved):
LGTM

PR #331, opened 2023-03-27
Fetch database list information from the canonical API endpoint

aris-aiven:
The database list information is currently extracted from the service results.
Change the implementation to use the canonical API endpoint, which has the best support for available service types out of the box.

The database list API endpoint is the canonical way to get the information.


alanfranz (dismissed):
Makes sense to me. Leaving open for aiven client people to double check.

fingon (approved):

PR #334, opened 2023-05-03
Add command to update oauth2 client [EH-441]

m-novikov:

To be able to edit existing oauth2 clients fixing improving title and descriptions


rsalevsky (approved):
Works

PR #337, opened 2023-06-16
FIX: Handle dict value for column in table_layout

bugant:
# Handle dict value for column in table_layout

Properly handle the case in which `table_layout` in `aiven.client.pretty.print_table` include a column whose value is a dict
object: in this case the dict will be serailized to a JSON string an added to the resulting table.

For example, consider the following:

```
rows = []
rows.append(
    {
        "ip": ipaddress.IPv4Address("192.168.16.1"),
        "network": ipaddress.IPv4Network("192.168.16.0/20"),
        "metric": {"qos": 1, "ping": 3},
        "next_hop_ip": ipaddress.IPv4Address("192.168.16.2"),
        "next_hop_mac": "0c:d0:f8:a3:04:31",
        "function1": test_print_table,
    }
)
rows.append(
    {
        "ip": ipaddress.IPv4Address("10.0.0.1"),
        "network": ipaddress.IPv4Network("10.0.0.0/16"),
        "metric": {"ping": 100},
        "function2": test_print_table,
    }
)
print_table(rows, table_layout=["ip", "metric", "network"])
```

The expected result would be:

```
IP            METRIC                 NETWORK
============  =====================  ===============
192.168.16.1  {"ping": 3, "qos": 1}  192.168.16.0/20
10.0.0.1      {"ping": 100}          10.0.0.0/16
```

At the moment, a `KeyError` will be raised, instead.  This PR fixes this case and makes the result matching the expected behaviour.

### DIFF
--- a/aiven/client/pretty.py
+++ b/aiven/client/pretty.py
@@ -131,11 +131,16 @@ def yield_table(  # noqa
         for key, value in item.items():
             if key in drop_fields:
                 continue  # field will not be printed
+            got_value_for_key = False
             for subkey, subvalue in iter_values(key, value):
                 if table_layout is not None and subkey not in flattened_table_layout:
                     continue  # table_layout has been specified but this field will not be printed
+                got_value_for_key = True
                 formatted_row[subkey] = format_item(subkey, subvalue)
                 widths[subkey] = max(len(subkey), len(formatted_row[subkey]), widths.get(subkey, 1))
+            if not got_value_for_key and (table_layout is None or key in flattened_table_layout):
+                formatted_row[key] = format_item(key, value)
+                widths[key] = max(len(key), len(formatted_row[key]), widths.get(key, 1))
 
     # default table layout is one row per item with sorted field names
     if table_layout is None:
--- a/tests/test_pretty.py
+++ b/tests/test_pretty.py
@@ -161,7 +161,7 @@ def test_print_table_nested_dict() -> None:
 
     actual = get_output(rows, table_layout=["ip", "metric", "network"])
     expected = """
-IP            METRIC                 NETWORK       
+IP            METRIC                 NETWORK
 ============  =====================  ===============
 192.168.16.1  {"ping": 3, "qos": 1}  192.168.16.0/20
 10.0.0.1      {"ping": 100}          10.0.0.0/16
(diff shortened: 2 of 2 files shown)


aiven/client/pretty.py
```diff
@@ -131,11 +131,16 @@ def iter_values(key: str, value: Any) -> Iterator[Tuple[str, Any]]:
...
+            got_value_for_key = False
             for subkey, subvalue in iter_values(key, value):
                 if table_layout is not None and subkey not in flattened_table_layout:
                     continue  # table_layout has been specified but this field will not be printed
+                got_value_for_key = True
                 formatted_row[subkey] = format_item(subkey, subvalue)
                 widths[subkey] = max(len(subkey), len(formatted_row[subkey]), widths.get(subkey, 1))
+            if not got_value_for_key and (table_layout is None or key in flattened_table_layout):
+                formatted_row[key] = format_item(key, value)
+                widths[key] = max(len(key), len(formatted_row[key]), widths.get(key, 1))
```
byashimov, line 143: I think this block is quite complicated already :( 
How about changing line 151 (old 146)?

```python
# Column names must have width
horizontal_fields: Collection[str] = [x for x in table_layout[0] if x in widths]
```
Though, accessing by index `[0]` looks weird, by that time `table_layout` must be a list of lists. It is casted on 144 (149) line.

bugant: If you do like that the supposedly missing column would not be shown, right?

I mean in that case if you do:

```
rows = []
rows.append(
    {
        "ip": ipaddress.IPv4Address("192.168.16.1"),
        "network": ipaddress.IPv4Network("192.168.16.0/20"),
        "metric": {"qos": 1, "ping": 3},
        "next_hop_ip": ipaddress.IPv4Address("192.168.16.2"),
        "next_hop_mac": "0c:d0:f8:a3:04:31",
        "function1": test_print_table,
    }
)
rows.append(
    {
        "ip": ipaddress.IPv4Address("10.0.0.1"),
        "network": ipaddress.IPv4Network("10.0.0.0/16"),
        "metric": {"ping": 100},
        "function2": test_print_table,
    }
)
print_table(rows, table_layout=["ip", "metric", "network"])
```

You will get:

```
IP            NETWORK
============  ===============
192.168.16.1  192.168.16.0/20
10.0.0.1      10.0.0.0/16
```

since the metrics column won't be in there, right?

byashimov: Oh. You are right. I've missed the whole thing. Please ignore me.

byashimov (approved):
lgtm

bugant:
Superseded by https://github.com/aiven/aiven-client/pull/340

PR #338, opened 2023-06-19
Add support for Opensearch Security Management API

mteiste:
Opensearch security plugin for the Aiven for Opensearch has been introduced with API to query the status and enable the functionality. This adds the matching commands to the client.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -3092,6 +3092,70 @@ ssl.truststore.type=JKS
 
     @arg.project
     @arg.service_name
+    def service__opensearch_security_management__status(self) -> None:
+        """Show status of the opensearch security"""
+        response = self.client.opensearch_security_get(
+            project=self.get_project(),
+            service=self.args.service_name,
+        )
+        available = response.get("security_plugin_available", False)
+        admin_enabled = response.get("security_plugin_admin_enabled", False)
+        message = response.get("message")
+        if not available:
+            if not message:
+                print("Opensearch Security is not available for the service")
+                return
+            print(message)
+            return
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -3090,6 +3090,70 @@ def service__es_acl_del(self) -> None:
...
+        )
+        available = response.get("security_plugin_available", False)
+        admin_enabled = response.get("security_plugin_admin_enabled", False)
+        message = response.get("message")
+        if not available:
+            if not message:
+                print("Opensearch Security is not available for the service")
+                return
+            print(message)
+            return
```
Prime541, line 3109: This could be simplified
```python
print(message or "Opensearch Security is not available for the service")
return
```

aiven/client/cli.py
```diff
@@ -3090,6 +3090,70 @@ def service__es_acl_del(self) -> None:
...
+
+    @arg.project
+    @arg.service_name
+    def service__opensearch_security_management__set(self) -> None:
+        """Set the password for the opensearch security management"""
+        print("Opensearch Security Management is enabled by setting the password")
+        print("for the security management user. Once enabled normal Aiven service")
+        print("user and ACL management is no longer used and all user and access control")
+        print("can only be done using the native Opensearch Security API or dashboard.")
+        if self.confirm("this action is unrevertable. Proceed (y/N)? "):
```
giacomo-alzetta-aiven, line 3123: nit:
```suggestion
        if self.confirm("this action cannot be reverted. Proceed (y/N)? "):
```

giacomo-alzetta-aiven (approved):

PR #341, opened 2023-06-22
add organization create command [EH-388]

watson28:
# add organization create command
New command for creating organizations. It receives the organization's name. The command asks for confirmation that the user understands what it means to create an Organization.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -1418,11 +1418,9 @@ class AivenCLI(argx.CommandLineTool):
         self.print_response(backups, json=self.args.json, table_layout=layout)
 
     def _get_project_ca(self) -> str:
-
         return self.client.get_project_ca(project=self.get_project())["certificate"]
 
     def _get_store_from_args(self) -> Store:
-
         if self.args.overwrite:
             store = Store.overwrite
         elif self.args.write:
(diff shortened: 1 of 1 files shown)


aiven/client/cli.py
```diff
@@ -5537,6 +5535,32 @@ def service__flink__cancel_application_deployments(self) -> None:
             ),
         )
 
+    @arg.json
+    @arg("-n", "--name", required=True, help="Name of the organization to create")
```
ngilles-aiven, line 5539: I would put this as an argument rather than an option (especially since it's `required=True`) ... it does feel like we have a bit of everything in this respect, but `service_name` is an argument as well.

ngilles-aiven (changes requested):
LGTM, just a point about arg vs option for the name

tests/test_argx.py
```diff
@@ -80,9 +80,9 @@ def test_extended_command_has_function_help() -> None:
 
     cli.extend_commands(sl)
 
-    help = cli.subparsers.choices[sl.bbb.__name__].format_help()
```
watson28, line 83: for some reason, this line started to rise the pylint error:
```
python3 -m pylint aiven tests
************* Module tests.test_argx
tests/test_argx.py:83:4: W0622: Redefining built-in 'help' (redefined-builtin)
```

dogukancagatay: It is fixed in #344 (https://github.com/aiven/aiven-client/pull/344/commits/e18d70f13e982a9776042926ac811b15c86d0f43)

rominf:
Please rebase on main, I've merged the fix for pylint (#345).

ngilles-aiven (approved):

PR #346, opened 2023-07-03
add user groups commands [EH-516]

dogukancagatay:
Adds user group commands list, show, create, update and delete.

[EH-516]

[EH-516]: https://aiven.atlassian.net/browse/EH-516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -40,6 +40,11 @@ AUTHENTICATION_METHOD_COLUMNS = [
     "create_time",
     "update_time",
 ]
+USER_GROUP_COLUMNS = [
+    "user_group_name",
+    "user_group_id",
+    "description",
+]
 EOL_ADVANCE_WARNING_TIME = timedelta(weeks=26)  # Give 6 months advance notice for EOL services
 
 
@@ -5575,7 +5580,7 @@ server_encryption_options:
         self.print_response(organizations, json=self.args.json, table_layout=layout)
 
     @arg.json
-    @arg.positional_organization_id
+    @arg.organization_id_positional
     @arg("-n", "--name", required=True, help="New name for the organization")
     def organization__update(self) -> None:
         """Update an organization"""
@@ -5589,7 +5594,7 @@ server_encryption_options:
         organization = self.client.update_organization(self.args.organization_id, self.args.name)
         self.print_response(organization, json=self.args.json, single_item=True, table_layout=layout)
 
-    @arg.positional_organization_id
+    @arg.organization_id_positional
     @arg.force
     def organization__delete(self) -> None:
         """Delete an organization"""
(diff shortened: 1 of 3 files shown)


aiven/client/cliarg.py
```diff
@@ -132,6 +132,7 @@ def wrapped(self: CommandLineTool) -> T:
     help="Minimum required nodes In Sync Replicas (ISR) to produce to a partition (default: 1)",
 )
 arg.organization_id = arg("--organization_id", help="Organization identifier")
+arg.positional_group_id = arg("group_id", help="Organization user group identifier")
```
ngilles-aiven, line 135: Arg names should use `-` instead of `_`

```suggestion
arg.positional_group_id = arg("group-id", help="Organization user group identifier")
```

aiven/client/cli.py
```diff
@@ -5587,6 +5590,83 @@ def organization__user__invite(self) -> None:
         """Invite user to join an organization"""
         self.client.invite_organization_user(self.args.organization_id, self.args.email)
 
+    @arg.json
+    @arg.organization_id
```
ngilles-aiven, line 5594: Not specifying an organization id crashes things as we are passing `None` down the stack. We should make `organization-id` option mandatory for these commands, I don't think we have anything else using this so we could just add `required=True` (or create a parallel one called `organization_id_mandatory`).

ngilles-aiven (changes requested):

aiven/client/cli.py
```diff
@@ -5626,6 +5631,83 @@ def organization__user__invite(self) -> None:
...
+            "user_group_name",
+            "description",
+            "user_group_id",
+            "member_count",
+        ]
+        self.print_response(groups, json=self.args.json, table_layout=layout)
+
+    @arg.json
+    @arg.organization_id
+    @arg.group_id_positional
```
ngilles-aiven, line 5649: looks like I was overzealous and actually broke things with using `-` in the `*_positional` arg definitions, could you revert those to `_` 🙏 ?

aiven/client/cli.py
```diff
@@ -40,6 +40,11 @@
     "create_time",
     "update_time",
 ]
+USER_GROUP_COLUMNS = [
+    "user_group_name",
+    "description",
+    "user_group_id",
```
ngilles-aiven, line 46: I feel somewhat biased to having the "primary key" be the leftmost column, as it's more easily findable/selectable by a human who'll likely copy paste it for the next command where they need to specify that value ... and second column be the friendly human readable name, cause ideally it would be close by for easy matching up?


```suggestion
    "user_group_id",
    "user_group_name",
    "description",
```

aiven/client/cli.py
```diff
@@ -5626,6 +5631,83 @@ def organization__user__invite(self) -> None:
...
+    @arg.json
+    @arg.organization_id
+    def organization__group__list(self) -> None:
+        """List user groups in an organization"""
+        groups = self.client.list_user_groups(self.args.organization_id)
+        layout = [
+            "user_group_name",
+            "description",
+            "user_group_id",
+            "member_count",
```
ngilles-aiven, line 5643: looks so close `USER_GROUP_COLUMS` almost feels like it could `USER_GROUP_COLUMNS + ['member_count']`  but that also feels like trying to be too DRY 🤷 

But at least same argument for ordering of colums applies.

ngilles-aiven (changes requested):

aiven/client/cli.py
dogukancagatay: Made `user_group_id` the second.

aiven/client/cli.py
dogukancagatay: Not anymore, took description to the very last

ngilles-aiven (approved):

PR #347, opened 2023-07-03
fix organization id args to use `-` instead of `_`

ngilles-aiven:
CLI args use `-` as word separator, some `_` slipped in.


dogukancagatay (approved):
LGTM

PR #349, opened 2023-07-06
Drop Python 3.7 support

rominf:

Python 3.7 recently reached its end-of-life:
https://devguide.python.org/versions/#versions

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -52,7 +52,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', 'pypy-3.7', 'pypy-3.8', 'pypy-3.9']
+        python-version: ['3.8', '3.9', '3.10', '3.11', 'pypy-3.8', 'pypy-3.9']
         os: [macos-latest, ubuntu-latest, windows-latest]
 
     steps:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,13 +13,12 @@ authors = [
 ]
 description="Aiven.io client library / command-line client"
 readme = "README.rst"
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 classifiers=[
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
     "Topic :: Software Development :: Libraries",
     "License :: OSI Approved :: Apache Software License",
-    "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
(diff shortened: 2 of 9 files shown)


bugant (approved):

bugant (approved):

.github/workflows/build.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', 'pypy-3.7', 'pypy-3.8', 'pypy-3.9']
+        python-version: ['3.8', '3.9', '3.10', '3.11', 'pypy-3.8', 'pypy-3.9']
```
Mulugruntz, line 55:
```suggestion
        python-version: ['3.8', '3.9', '3.10', '3.11', 'pypy-3.9', 'pypy-3.10']
```

It seems that officially, pypy is on 3.9 and 3.10: https://www.pypy.org/download_advanced.html

We could also decide for pypy to leave 3.8 and add 3.10 (3.8, 3.9, 3.10).

pyproject.toml
```diff
@@ -66,7 +65,7 @@ version-file = "aiven/client/version.py"
 
 [tool.black]
 line-length = 125
-target-version = ['py37', 'py38', 'py39']  # 'py310' is not yet available on Fedora 34
+target-version = ['py38', 'py39']  # 'py310' is not yet available on Fedora 34
```
Mulugruntz, line 68: Not Fedora34 anymore.
Can add
```suggestion
target-version = ['py38', 'py39', 'py310', 'py311']
```

.github/workflows/build.yaml
rominf: Thanks, I noticed that. I wanted to put this into separate PR to make PR and commit history clean.

pyproject.toml
rominf: Thanks, I noticed that. I wanted to put this into separate PR to make PR and commit history clean.

Mulugruntz: As much as the pypy line could make sense to have in a separate PR, I think this one here is part of this PR.
You can amend + force push to have a clean history.

rominf: I still think that it should go into separate PR, as it requires bumping requirements in `pyproject.toml` file (`black==23.3.0`). It's better than both: PR and commits are atomic, it makes easier to reason about the changes in the future.

Mulugruntz: Okay, let's do that in another PR then!

Mulugruntz (approved):

Mulugruntz:
@rominf I can merge, but ideally, it'd be great to have an internal ticket number and put it in the PR description beforehand.
WDYT?

ngilles-aiven (approved):

ngilles-aiven:
@Mulugruntz oops, didn't see your last comment there. I don't think it's super important to have an internal Jira number in this case... in fact aiven-client is public project (we allow external contributors), and we used to not put internal numbers in there at all. At some point we gave up on that and added them anyway for Jira integration to work.

PR #351, opened 2023-07-07
Add ruff, drop pylint, upgrade syntax

rominf:

- Add ruff
- Drop pylint
- Upgrade syntax to Python 3.8

See more details in the commit messages.


aiven/client/argx.py
```diff
@@ -167,13 +150,13 @@ def save(self) -> None:
             jsonlib.dump(self, fp, sort_keys=True, indent=4)
 
 
-class CommandLineTool:  # pylint: disable=old-style-class
+class CommandLineTool:
     config: Config
 
-    def __init__(self, name: str, parser: Optional[argparse.ArgumentParser] = None):
+    def __init__(self, name: str, parser: argparse.ArgumentParser | None = None):
```
ngilles-aiven, line 156: `T1 | T2`  syntax is only in Python 3.10+ ... we can't get rid of it yet with 3.8 + 3.9 support :-(

aiven/client/argx.py
```diff
@@ -124,7 +107,7 @@ def __setattr__(self, name: str, value: Callable) -> None:
 arg = Arg()
 
 
-def name_to_cmd_parts(name: str) -> List[str]:
+def name_to_cmd_parts(name: str) -> list[str]:
```
ngilles-aiven, line 110: `list[T]` is only available in 3.9+

ngilles-aiven (changes requested):
I was all excited for this,  but then realised/remembered that the `list[T]`l and `T1 | T2` are only available in 3.9 and 3.10 respectively, but we need to keep 3.8 compat 😞

aiven/client/argx.py
rominf: Starting with Python 3.7, when from __future__ import annotations is used, function and variable annotations can parameterize standard collections directly.

From:
https://peps.python.org/pep-0585/#implementation

aiven/client/argx.py
rominf: It's possible because of https://peps.python.org/pep-0563/. Also, we have testing suite which pass.
image

aiven/client/argx.py
rominf: https://github.com/python/mypy/issues/7907

aiven/client/argx.py
rominf: https://github.com/python/mypy/issues/9610

ngilles-aiven: I guess it works if we're using "recent enough" version of `mypy` ... and not using something like pydantic that actually looks and tries to resolve the type hints, or else it would result in something this:
```python
>>> from aiven.client.argx import name_to_cmd_parts
>>> from typing import get_type_hints
>>> get_type_hints(name_to_cmd_parts)
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.8/typing.py", line 1264, in get_type_hints
    value = _eval_type(value, globalns, localns)
  File "/usr/lib64/python3.8/typing.py", line 270, in _eval_type
    return t._evaluate(globalns, localns)
  File "/usr/lib64/python3.8/typing.py", line 518, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
TypeError: 'type' object is not subscriptable

Comment (rominf) [reply aiven/client/argx.py:156]: > I guess it works if we're using "recent enough" version of mypy

Mypy version pinned in dependencies supports this, so I see no issue here.

> and not using something like pydantic that actually looks and tries to resolve the type hints

I believe it's not that important.

File: aiven/client/cli.py [hunk 3]
```diff
@@ -138,10 +141,10 @@ def get_current_date() -> datetime:
 
 
 if (sys.version_info.major, sys.version_info.minor) >= (3, 8):
-    from typing import Protocol  # pylint: disable=no-name-in-module
+    from typing import Protocol
 
-    class ClientFactory(Protocol):  # pylint: disable=too-few-public-methods
-        def __call__(self, base_url: str, show_http: bool, request_timeout: Optional[int]) -> client.AivenClient:
+    class ClientFactory(Protocol):
+        def __call__(self, base_url: str, show_http: bool, request_timeout: int | None) -> client.AivenClient:
```
ngilles-aiven, line 147: nit: we can remove the (now redundant)`if` guard here which is for 3.8. I thought `pyupgrade` took care of this pattern, maybe `ruff`'s pyupgrade rules don't ?

ngilles-aiven (dismissed):
LGTM :-) just one minor nit

ngilles-aiven (approved):
🎉

PR #354, opened 2023-07-13
Fix get_stripe_key typing and handling

rominf:

Typing was broken. This caused mypy checking errors downstream.

### DIFF
--- a/aiven/client/client.py
+++ b/aiven/client/client.py
@@ -1829,7 +1829,7 @@ class AivenClient(AivenClientBase):
     def remove_card(self, card_id: str) -> Mapping:
         return self.verify(self.delete, self.build_path("card", card_id))
 
-    def get_stripe_key(self) -> str:
+    def get_stripe_key(self) -> Mapping[str, str]:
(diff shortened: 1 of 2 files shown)


aiven/client/client.py
```diff
@@ -1829,7 +1829,7 @@ def update_card(self, card_id: str, **kwargs: Any) -> Mapping:
     def remove_card(self, card_id: str) -> Mapping:
         return self.verify(self.delete, self.build_path("card", card_id))
 
-    def get_stripe_key(self) -> str:
+    def get_stripe_key(self) -> Mapping[str, str]:
         return self.verify(self.get, self.build_path("config", "stripe_key"), result_key="stripe_key")
```
dogukancagatay, line 1833: We should remove the `result_key` parameter, otherwise it won't return the map with `stripe_key` item.

```suggestion
        return self.verify(self.get, self.build_path("config", "stripe_key"))
```

dogukancagatay (changes requested):
We need to remove the result_key to get the dict.

PR #356, opened 2023-07-21
Add clickhouse database create/delete/list

kmichel-aiven:

Make these operations available from the CLI.


alanfranz (approved):
LGTM

PR #361, opened 2023-09-12
Improve tests for Kafka topic create and update

ivanyu:


tests/test_cli.py
```diff
@@ -43,48 +46,128 @@ def test_service_user_create() -> None:
...
+    cli = build_aiven_cli(client)
+    assert cli.run(args=command_line.split(" ")) is None
+    session.post.assert_called_once_with(
+        "/v1/project/project1/service/service1/topic",
+        headers=ANY,
+        params=ANY,
+        data=ANY,  # checked below
     )
+    data_dict = json.loads(session.post.call_args_list[0].kwargs["data"])
+    assert data_dict == expected_post_data
```
eliax1996, line 114:
```suggestion
    client.session = session
    cli = build_aiven_cli(client)
    assert cli.run(args=command_line.split(" ")) is None
    data_dict = session.put.call_args_list[0].kwargs["data"]
    session.post.assert_called_once_with(
        "/v1/project/project1/service/service1/topic/topic1",
        headers=ANY,
        params=ANY,
        data=data_dict,
    )
```

maybe like that?

tests/test_cli.py
```diff
@@ -43,48 +46,128 @@ def test_service_user_create() -> None:
...
+    cli = build_aiven_cli(client)
+    assert cli.run(args=command_line.split(" ")) is None
+    session.put.assert_called_once_with(
+        "/v1/project/project1/service/service1/topic/topic1",
+        headers=ANY,
+        params=ANY,
+        data=ANY,  # checked below
     )
+    data_dict = json.loads(session.put.call_args_list[0].kwargs["data"])
+    assert data_dict == expected_put_data
```
eliax1996, line 170:
```suggestion
    client.session = session
    cli = build_aiven_cli(client)
    assert cli.run(args=command_line.split(" ")) is None
    data_dict = session.put.call_args_list[0].kwargs["data"]
    session.post.assert_called_once_with(
        "/v1/project/project1/service/service1/topic/topic1",
        headers=ANY,
        params=ANY,
        data=data_dict,
    )
```

eliax1996 (commented):
Syntattically it's fine, I'm asking a couple of question to understand also the semantics.

Are we executing similar tests in an end to end scenario against a real system?

Because they seems like a contract testing about the shape of the expected data. But, if we don't test that the provider is compliant with this specification, both the systems can succeed with the test suite but in reality they are talking a different language when deployed together they can fail.

Also here we have a dynamicity in the shape inside the test. We are not enforcing that the reply is of a certain shape but we are only enforcing that is containing exactly the same data that we have given to the other component, maybe we want also to have a set of keys that we expect always to have?

ivanyu:
> Are we executing similar tests in an end to end scenario against a real system?
>
> Because they seems like a contract testing about the shape of the expected data. But, if we don't test that the provider is compliant with this specification, both the systems can succeed with the test suite but in reality they are talking a different language when deployed together they can fail.

No, we don't have end-to-end tests for this client (except for some internal Aiven testing). Most tests in `test_cli.py` don't test much beyond the CLI parsing. What I added are indeed contract tests: we "know" the Aiven API is correct, so we're testing the client is compliant with its schema.

ivanyu:
> Also here we have a dynamicity in the shape inside the test. We are not enforcing that the reply is of a certain shape but we are only enforcing that is containing exactly the same data that we have given to the other component, maybe we want also to have a set of keys that we expect always to have?

I don't understand this. Could you elaborate?

eliax1996:
> No, we don't have end-to-end tests for this client (except for some internal Aiven testing). Most tests in test_cli.py don't test much beyond the CLI parsing. What I added are indeed contract tests: we "know" the Aiven API is correct, so we're testing the client is compliant with its schema.

Now I got everything, sorry for not understanding it initially!
LGTM

PR #362, opened 2023-09-15
Support setting Kafka topic retention in milliseconds

ivanyu:
The client doesn't resolve the priorities if both `--replication` and `--retention-ms` are provided, the API does (in favor of the latter).


eliax1996 (approved):
LGTM

PR #363, opened 2023-09-15
Support setting Kafka tiered storage topic configs

ivanyu:
`remote.storage.enable`, `local.retention.ms`, `local.retention.bytes`.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -2880,6 +2880,10 @@ ssl.truststore.type=JKS
     @arg.retention
     @arg.retention_ms
     @arg.retention_bytes
+    @arg.remote_storage_enable
+    @arg.remote_storage_disable
+    @arg.local_retention_ms
+    @arg.local_retention_bytes
     @arg.tag
     @arg(
         "--cleanup-policy",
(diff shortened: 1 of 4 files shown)


aiven/client/cli.py
```diff
@@ -2952,10 +2963,23 @@ def service__topic_update(self) -> None:
...
+            remote_storage_enable=self._remote_storage_enable(),
+            local_retention_ms=self.args.local_retention_ms,
+            local_retention_bytes=self.args.local_retention_bytes,
             tags=tags,
         )
         print(response["message"])
 
+    def _remote_storage_enable(self) -> bool | None:
+        if self.args.remote_storage_enable and self.args.remote_storage_disable:
+            raise argx.UserError("Only set at most one of --remote-storage-enable and --remote-storage-disable")
```
eliax1996, line 2975: Schrödinger remote storage mode on 😄

eliax1996: In this case probably isn't worth but we could always add a param to the annotation to do something like the following in case of N mutually exclusive options:
```
group = parser.add_mutually_exclusive_group()
group.add_argument('-a', action='store_true')
group.add_argument('-b', action='store_true')
```

eliax1996 (approved):
LGTM, are negative numbers a problem? Should we guard against that?

ivanyu:
> are negative numbers a problem? Should we guard against that?

I think not. `-1` and `-2` are valid and we rely on the backend to guard against the others.

PR #368, opened 2023-10-25
Add custom files API support

van-ess0:
This commit adds custom file support for Aiven services

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -5546,6 +5546,77 @@ server_encryption_options:
             ),
         )
 
+    @arg.project
+    @arg.service_name
+    def service__custom_file__list(self) -> None:
+        """List all the custom files for the specified service"""
+        self.print_response(
+            self.client.custom_file_list(
+                project=self.args.project,
+                service=self.args.service_name,
+            ),
+        )
+
+    @arg.project
+    @arg.service_name
+    @arg("--file_id", help="A file ID, usually an uuid4 string", required=True)
+    @arg("--target_filepath", help="Where to save the file downloaded")
+    @arg("--stdout_write", help="Write to the stdout instead of file", action="store_true")
+    def service__custom_file__get(self) -> None:
+        """
+        Download the custom file for the specified service and ID
+        Requires at least `--target_filepath` or `--stdout_write`, but able to handle both
+        """
+        if not (self.args.target_filepath or self.args.stdout_write):
+            raise argx.UserError("You need to specify `--target_filepath` or `--stdout_write` or both")
+
+        result = self.client.custom_file_get(
+            project=self.args.project,
+            service=self.args.service_name,
+            file_id=self.args.file_id,
+        )
+
+        if self.args.target_filepath:
+            with open(self.args.target_filepath, "wb") as f:
+                f.write(result)
+
+        if self.args.stdout_write:
+            print(result.decode("utf-8"))
+
+    @arg.project
+    @arg.service_name
+    @arg("--file_path", help="A path to the file", required=True)
+    @arg("--file_type", choices=["synonyms", "stopwords", "wordnet"], required=True)
+    @arg("--file_name", help="A name for the file", required=True)
+    def service__custom_file__upload(self) -> None:
+        """Upload custom file for the specified service"""
+        with open(self.args.file_path, "rb") as f:
+            self.print_response(
+                self.client.custom_file_upload(
+                    project=self.args.project,
+                    service=self.args.service_name,
+                    file_type=self.args.file_type,
+                    file_name=self.args.file_name,
+                    file_object=f,
+                ),
+            )
+
+    @arg.project
+    @arg.service_name
+    @arg("--file_path", help="A path to the file", required=True)
+    @arg("--file_id", help="A file ID, usually an uuid4 string", required=True)
+    def service__custom_file__update(self) -> None:
+        """Update custom file for the specified service and ID"""
+        with open(self.args.file_path, "rb") as f:
+            self.print_response(
+                self.client.custom_file_update(
+                    project=self.args.project,
+                    service=self.args.service_name,
+                    file_id=self.args.file_id,
+                    file_object=f,
+                ),
+            )
+
     @arg.json
     @arg("name", help="Name of the organization to create")
     @arg.force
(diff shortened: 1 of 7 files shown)


aiven/client/cli.py
```diff
@@ -5666,6 +5666,68 @@ def service__flink__cancel_application_deployments(self) -> None:
...
+                project=self.args.project,
+                service=self.args.service_name,
+            ),
+        )
+
+    @arg.project
+    @arg.service_name
+    @arg("--file_id", help="A file ID, usually an uuid4 string", required=True)
+    def service__custom_file__get(self) -> None:
+        """List all the custom files for the specified service"""
```
mteiste, line 5684: The comment does not match the function

aiven/client/cli.py
```diff
@@ -5666,6 +5666,68 @@ def service__flink__cancel_application_deployments(self) -> None:
...
+            ),
+        )
+
+    @arg.project
+    @arg.service_name
+    @arg("--file_path", help="A path to the file", required=True)
+    @arg("--file_type", choices=["synonyms", "stopwords"], required=True)
+    @arg("--file_name", help="A name for the file", required=True)
+    def service__custom_file__upload(self) -> None:
+        """List all the custom files for the specified service"""
```
mteiste, line 5699: Description does not match the function

aiven/client/cli.py
```diff
@@ -5666,6 +5666,68 @@ def service__flink__cancel_application_deployments(self) -> None:
...
+                ),
+            )
+
+    @arg.project
+    @arg.service_name
+    @arg("--file_path", help="A path to the file", required=True)
+    @arg("--file_type", choices=["synonyms", "stopwords"], required=True)
+    @arg("--file_name", help="A name for the file", required=True)
+    @arg("--file_id", help="A file ID, usually an uuid4 string", required=True)
+    def service__custom_file__update(self) -> None:
```
mteiste, line 5717: Description does not match the function

mteiste (dismissed):

aiven/client/client.py
```diff
@@ -174,11 +179,28 @@ def verify(
                 )
                 time.sleep(0.2)
 
+        return self._process_response(response=response, op=op, path=path, result_key=result_key)
+
+    @staticmethod
+    def build_path(*parts: str) -> str:
+        return "/" + "/".join(quote(part, safe="") for part in parts)
```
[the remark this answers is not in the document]
van-ess0: I don't think it worth it here. I've always considered `os.path.join()` as an incapsulation of building path for different OSes: choosing `/` vs `\`, setting correct root like `C:` vs `/`, etc. Also since this util is supposed to be used on Windows we may face some bugs related to it.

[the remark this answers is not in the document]
heikju: I agree here, `os.path.join()` is specifically for concatenating file paths using operator system specific conventions. HTTP request URLs are a different thing, and the ways these two are built don't necessarily match.

aiven/client/client.py
```diff
@@ -174,11 +179,28 @@ def verify(
...
+        path: str,
+        result_key: str | None = None,
+    ) -> Mapping | str:
         # Check API is actually returning data or not
         if response.status_code == HTTPStatus.NO_CONTENT or response.headers.get("Content-Length", "0") == "0":
             return {}
 
-        result = response.json()
+        try:
+            result = response.json()
```
heikju, line 200: I think a choice between decoding and not decoding this should be done based on response `Content-Type` so that 1. we don't decode and thus modify something that just happens to look like JSON if it's not intended to be processed as such, and 2. if backend malfunctions and returns broken or incomplete JSON, we can fail instead of passing broken data to e.g. another process in a shell pipeline that may not be prepared for such.

Especially if there's custom file data as `application/octet-stream`, I wouldn't expect client to decode it and output a possibly modified version of the contents, even if it happens to look like JSON.

aiven/client/cli.py
```diff
@@ -5666,6 +5666,64 @@ def service__flink__cancel_application_deployments(self) -> None:
...
+                service=self.args.service_name,
+            ),
+        )
+
+    @arg.project
+    @arg.service_name
+    @arg("--file_id", help="A file ID, usually an uuid4 string", required=True)
+    def service__custom_file__get(self) -> None:
+        """View the custom file for the specified service and ID"""
+        print(
```
heikju, line 5685: Looking at the upload implementation in the client, the custom files are `application/octet-stream` i.e. binary data, does `print()` always write to stdout such data exactly as it, bit-by-bit?

And with binary data, an option to output to a file could be a convenient option (though one can do a shell redirect too of course).

heikju (changes requested):

heikju:
It would also be nice to see tests for this. The test coverage for client is not great, but at least we would have an opportunity here to make it slightly better.

aiven/client/cli.py
van-ess0: Since we assume cutom files to be utf-8 encoded text files, it should be save to simply redirect the output of `print`

rikonen: Custom files going forward can be anything. I would not have any implementation that prints to stdout but require target file name and write there. Some special syntax to allow using stdout as the target file could be allowed.

aiven/client/client.py
van-ess0: Makes sense, changed it, thank you

aiven/client/cli.py
```diff
@@ -5546,6 +5546,77 @@ def service__flink__cancel_application_deployments(self) -> None:
...
+                project=self.args.project,
+                service=self.args.service_name,
+            ),
+        )
+
+    @arg.project
+    @arg.service_name
+    @arg("--file_id", help="A file ID, usually an uuid4 string", required=True)
+    @arg("--target_filepath", help="Where to save the file downloaded")
+    @arg("--stdout_write", help="Write to the stdout instead of file", action="store_true")
```
mteiste, line 5564: Not sure if this flag makes sense. E.g from the face of it (if I read this correctly) the default if target_filepath is not given is to write to stdout, so the question is what is the use case for getting the file but not writing it to anywhere?

mteiste (changes requested):

aiven/client/cli.py
van-ess0: I see it like `fetch file | grep foo | sed 's/o/i/' >> new_file.txt`

mteiste (approved):

PR #369, opened 2023-11-03
Fix API client compliance with HTTP/1.1

sigmaris:

In HTTP/1.1 Content-Length is not always required to be sent by the server - if chunked transfer encoding is used, the Content-Length header can be omitted, and instead length indicated as part of each chunk. Chunked transfer encoding is [always acceptable in HTTP/1.1][1] so by effectively requiring a Content-Length header we are breaking compliance with the spec.

We have avoided this biting us, purely by accident so far due to the web server used for our API, but it will cause problems in the future. Basically, we shouldn't try and figure out the content length by parsing headers in our code, but instead rely on the Requests library which has logic to calculate the correct value based on the HTTP spec.

[1]: https://datatracker.ietf.org/doc/html/rfc9112#section-7.4-2

### DIFF
--- a/aiven/client/client.py
+++ b/aiven/client/client.py
@@ -175,7 +175,7 @@ class AivenClientBase:
                 time.sleep(0.2)
 
         # Check API is actually returning data or not
-        if response.status_code == HTTPStatus.NO_CONTENT or response.headers.get("Content-Length", "0") == "0":
+        if response.status_code == HTTPStatus.NO_CONTENT or len(response.content) == 0:
             return {}
 
         result = response.json()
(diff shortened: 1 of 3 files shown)


aiven/client/client.py
```diff
@@ -175,7 +175,7 @@ def verify(
                 time.sleep(0.2)
 
         # Check API is actually returning data or not
-        if response.status_code == HTTPStatus.NO_CONTENT or response.headers.get("Content-Length", "0") == "0":
+        if response.status_code == HTTPStatus.NO_CONTENT or len(response.content) == 0:
```
giacomo-alzetta-aiven, line 178: content is always read just below so I guess this doesn't really introduce any performance penalty compared to the previous implementation.

giacomo-alzetta-aiven (approved):
LGTM

PR #374, opened 2023-11-28
Initial commands on the sustainability APIs

hnousiainen:
Add two new commands to assess carbon impact of Aiven services. The two commands are:
1. Command to project or predict typical emissions for a given service type, plan and cloud selection, and
2. Command to assess project actual emissions over a time period

The backend reporting functionality is still considered experimental, and we're working to ensure validation on the returned numbers.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -5807,6 +5807,24 @@ server_encryption_options:
 
         self.print_response(records, json=self.args.json, table_layout=layout)
 
+    @arg.project
+    @arg.json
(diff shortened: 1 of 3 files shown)


giacomo-alzetta-aiven (changes requested):
Sorry for the long delay in looking at this PR.

PR looks good, but there are conflicts. If you can update it I will merge it asap.

giacomo-alzetta-aiven (approved):
LGTM

PR #377, opened 2024-01-16
Alphabetically list any `--options` printed by `--help`.

neilhwatson:
Longer lists of options are difficult for humans to read when they are unsorted.

As an Aiven operator I find it difficult to read through long lists of unsorted help options. I hope this will help existing a future operators and user.


dogukancagatay (approved):
LGTM

PR #384, opened 2024-05-29
feature: Expose retry paramater for update_service_topic

aiven-anton:

To allow controlling retry behavior, this commit adds the ability to pass in a `RetrySpec` when instantiating `AivenClient`. This maintains existing behavior for all current instantiations, but when configured this overrides these properties of all calls made through the client:

- Passing `RetrySpec(http_methods=("GET", "PUT"))` will make the client use retries for PUT in addition to the default GET.
- Passing `RetrySpec(sleep=timedelta(seconds=1))` changes sleep time in between retries to 1 second for GET calls, other methods will have no retries.
- Passing `RetrySpec(attempts=5)` changes the number of attempts for GET calls, other methods will have no retries.
- Passing `RetrySpec(attempts=5, sleep=timedelta(seconds=1), http_methods=("GET", "PUT"))` will set the configured values for GET and PUT operations, all other methods will have no retries.

In short, while maintaining existing behavior, this commit introduces ability to configure which methods should use retries, and it allows configuring how many and the sleep time in between.

### DIFF
--- a/aiven/client/_typing.py
+++ b/aiven/client/_typing.py
@@ -1,0 +1,10 @@
+from typing import NoReturn
+
+
+def assert_never(arg: NoReturn, /) -> NoReturn:
+    """
+    Backport of standard library typing.assert_never.
+
+    https://docs.python.org/3/library/typing.html#typing.assert_never
+    """
+    raise AssertionError("Expected code to be unreachable")
--- a/aiven/client/client.py
+++ b/aiven/client/client.py
@@ -4,14 +4,28 @@
 # See the file `LICENSE` for details.
 from __future__ import annotations
 
+from ._typing import assert_never
 from .common import UNDEFINED
 from .session import get_requests_session
 from http import HTTPStatus
 from requests import Response
 from requests_toolbelt import MultipartEncoder  # type: ignore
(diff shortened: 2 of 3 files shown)


aiven-anton:
@rikonen That should probably only be applied for the typical idempotent verbs? I.e. `PUT` and `PATCH`, but not `POST`?

rikonen:
Possibly, or then just allow defining the verbs with those being the default when nothing else is specified.

aiven-anton:
Changing to introduce retry behavior everywhere we currently don't seems scary, are we confident all existing `PUT` and `PATCH` are idempotent?

rikonen:
There are too many to say that for certain, but the default behavior would not change and this would be conscious decision by the caller. In some scenarios like when running tests over the API it may not matter even if in some rare case the behavior would be unexpected.

aiven-anton:
@rikonen I decided to attempt going a bit further in allowing control of retry behavior. This still maintains exactly the existing behavior when `AivenClient` is instantiated without a `default_retry_spec`. When it is passed, it now allows controlling all of:

- Which HTTP methods exercises retry behavior.
- Sleep time in between retries.
- Number of retry attempts.

Let me know if this makes sense.

rikonen (approved):
Looks good to me. Leaving still open in case someone else wants to have a look.

PR #385, opened 2024-06-07
client: allow to provision Google BYOC [MILK-135]

Prime541:
In order to provision a Google BYOC we need to pass the customer's privilege-bearing service account ID, so we can later impersonate it and start service node instances in the Google account of the customer.

Also, ease local testing:
- replace 'pip install -e .[dev]' by 'make install-py'
- create a default 'make' target to locally run: install-py, validate-style, lint and test

[MILK-135]

[MILK-135]: https://aiven.atlassian.net/browse/MILK-135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aiven/client/client.py
+++ b/aiven/client/client.py
@@ -1,4 +1,4 @@
-# Copright 2015, Aiven, https://aiven.io/
+# Copyright 2015, Aiven, https://aiven.io/
 #
 # This file is under the Apache License, Version 2.0.
 # See the file `LICENSE` for details.
--- a/aiven/client/client.py
+++ b/aiven/client/client.py
@@ -1,4 +1,4 @@
-# Copyright 2015, Aiven, https://aiven.io/
+# Copright 2015, Aiven, https://aiven.io/
 #
 # This file is under the Apache License, Version 2.0.
 # See the file `LICENSE` for details.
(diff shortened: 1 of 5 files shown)


Prime541:
Moving back to `Draft` since this change should not be public yet.

aiven/client/client.py
```diff
@@ -2750,9 +2750,18 @@ def byoc_update(
...
-    def byoc_provision(self, *, organization_id: str, byoc_id: str, aws_iam_role_arn: str | None) -> Mapping[Any, Any]:
+    def byoc_provision(
+        self,
+        *,
+        organization_id: str,
+        byoc_id: str,
+        aws_iam_role_arn: str | None = None,
+        google_privilege_bearing_service_account_id: str | None = None,
+    ) -> Mapping[Any, Any]:
         if aws_iam_role_arn is not None:
```
allaouiamine, line 2761: if both `aws_iam_role_arn` and `google_privilege_bearing_service_account_id` are used, we will try to provision AWS byoc. This behaviour might get unpredictable once we support even more cloud providers

Can we add a check before to make sure there exactly one of these IAM role like arguments? what do you think?

heikju (approved):

PR #386, opened 2024-07-04
Add support for Valkey ACL users

bogdanp05:

With the release of [Aiven for Valkey](https://aiven.io/docs/products/valkey), it's possible to create Valkey ACL users via the API. 
- enable Valkey ACL user creation and editing via the aiven client
- replace Redis ACL user example with Valkey
- fix `make ruff` recipe


giacomo-alzetta-aiven (approved):
LGTM

PR #389, opened 2024-10-02
Support alloydbomni service type

ettanany:
With this change, we ensure existing PG commands also work for AlloyDB Omni.


alexole (approved):
LGTM

PR #391, opened 2024-11-13
client: flink commands ignore the current project

jclarysse:
Customers are used to set environment so they don't have to pass all arguments for each aiven-client command they run. This PR takes the current project into account in flink commands.

[PM-4549]

[PM-4549]: https://aiven.atlassian.net/browse/PM-4549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -5349,7 +5349,7 @@ server_encryption_options:
         """List Flink applications"""
         self.print_response(
             self.client.flink_list_applications(
-                project=self.args.project,
+                project=self.get_project(),
                 service=self.args.service_name,
             ),
         )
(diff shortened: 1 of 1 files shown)


gpkc (approved):

jclarysse:
> we should fix the get_project to reflect the correct hierarchy of inference which should be argument -> environment -> config.
> we should fix the args.project default to behave similarly so as not to mislead developers in the future.
@dogukancagatay Do we agree that this should be a separate ticket? Thanks

PR #393, opened 2024-11-18
Add cli support for Apache Kafka native ACLs

biggusdonzus:
Add the following `service` subcomands `kafka-acl-list` `kafka-acl-add` `kafka-acl-delete`.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -2954,7 +2954,7 @@ ssl.truststore.type=JKS
         required=True,
     )
     def service__acl_add(self) -> None:
-        """Add a Kafka ACL entry"""
+        """Add an Aiven ACL for Kafka entry"""
         response = self.client.add_service_kafka_acl(
             project=self.get_project(),
             service=self.args.service_name,
@@ -2968,7 +2968,7 @@ ssl.truststore.type=JKS
     @arg.service_name
     @arg("acl_id", help="ID of the ACL entry to delete")
     def service__acl_delete(self) -> None:
-        """Delete a Kafka ACL entry"""
+        """Delete an Aiven ACL for Kafka entry"""
         response = self.client.delete_service_kafka_acl(
             project=self.get_project(), service=self.args.service_name, acl_id=self.args.acl_id
         )
@@ -2978,7 +2978,7 @@ ssl.truststore.type=JKS
     @arg.service_name
     @arg.json
     def service__acl_list(self) -> None:
-        """List Kafka ACL entries"""
+        """List Aiven ACL for Kafka entries"""
         service = self.client.get_service(project=self.get_project(), service=self.args.service_name)
 
         layout = ["id", "username", "topic", "permission"]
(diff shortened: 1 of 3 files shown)


aiven/client/cli.py
```diff
@@ -6176,6 +6176,104 @@ def service__alloydbomni__google_cloud_private_key__delete(self) -> None:
...
+            "a name of '*' matches all resources of the selected type"
+        ),
+        required=True,
+    )
+    @arg(
+        "--resource-type",
+        help="Topic resource type to which ACLs should be added",
+        required=False,
+        choices=["Any", "Topic", "Group", "Cluster", "TransactionalId", "DelegationToken"],
+    )
```
tvainika, line 6214: Just for user convenience, should we model command line args more similar to kafka-acls.sh?

It would mean that instead of simply giving resource-name as parameter, usage would be more like
`[--cluster | --group group | --topic topic | --transactional-id transactional-id ] `
where only one of these is given.

Also for cluster type name must be always `kafka-cluster`.

Also we could skip DelegationToken here because there is currently no support for those in aiven services.

aiven/client/cli.py
```diff
@@ -6176,6 +6176,104 @@ def service__alloydbomni__google_cloud_private_key__delete(self) -> None:
...
+            operation=self.args.operation,
+            permission_type=self.args.permission_type,
+        )
+        print(response["message"])
+
+    @arg.project
+    @arg.service_name
+    @arg.json
+    def service__kafka_acl_list(self) -> None:
+        """List Kafka native ACL entries"""
```
tvainika, line 6258:
```suggestion
        """List Kafka-native ACL entries"""
```

aiven/client/cli.py
```diff
@@ -6176,6 +6176,104 @@ def service__alloydbomni__google_cloud_private_key__delete(self) -> None:
...
+        )
+        acls = response.get("kafka_acl", [])
+        layout = ["id", "principal", "operation", "resourceName", "resourceType", "patternType", "permissionType"]
+        self.print_response(acls, json=self.args.json, table_layout=layout)
+
+    @arg.project
+    @arg.service_name
+    @arg("acl_id", help="ID of the ACL entry to delete")
+    def service__kafka_acl_delete(self) -> None:
+        """Delete a Kafka ACL entry"""
```
tvainika, line 6271:
```suggestion
        """Delete a Kafka-native ACL entry"""
```

aiven/client/cli.py
```diff
@@ -6176,6 +6176,104 @@ def service__alloydbomni__google_cloud_private_key__delete(self) -> None:
...
+        required=False,
+        default="*",
+    )
+    @arg(
+        "--principal",
+        help="The principal for the ACLs, must be in the form principalType:name",
+        required=True,
+    )
+    def service__kafka_acl_add(self) -> None:
+        """Add a Kafka native ACL entry"""
```
tvainika, line 6240:
```suggestion
        """Add a Kafka-native ACL entry"""
```

aiven/client/cli.py
```diff
@@ -6176,6 +6176,104 @@ def service__alloydbomni__google_cloud_private_key__delete(self) -> None:
...
+    )
+    @arg(
+        "--permission-type",
+        help="The type of the resource pattern",
+        required=True,
+        choices=["ALLOW", "DENY"],
+    )
+    @arg(
+        "--host",
+        help="The host for the ACLs, a value of '*' matched all hosts",
```
tvainika, line 6230:
```suggestion
        help="The host for the ACLs, a value of '*' matches all hosts",
```

aiven/client/cli.py
```diff
@@ -6176,6 +6176,104 @@ def service__alloydbomni__google_cloud_private_key__delete(self) -> None:
...
+        help="The type of the resource pattern",
+        required=False,
+        choices=["LITERAL", "PREFIXED"],
+        default="LITERAL",
+    )
+    @arg(
+        "--permission-type",
+        help="The type of the resource pattern",
+        required=True,
+        choices=["ALLOW", "DENY"],
```
tvainika, line 6226: More convenient if Allow is default?

tvainika (commented):
In addition existing Aiven ACLs for Kafka need doc polishing to make it clear what is where (use `avn service --help` to see those existing commands)

aiven/client/cli.py
```diff
@@ -6176,6 +6176,104 @@ def service__alloydbomni__google_cloud_private_key__delete(self) -> None:
...
+    @arg.service_name
+    @arg.json
+    def service__kafka_acl_list(self) -> None:
+        """List Kafka native ACL entries"""
+        response = self.client.service_kafka_native_acl_list(
+            project=self.get_project(),
+            service=self.args.service_name,
+        )
+        acls = response.get("kafka_acl", [])
+        layout = ["id", "principal", "operation", "resourceName", "resourceType", "patternType", "permissionType"]
```
tvainika, line 6264: This is missing host. For convenience it could be so that host field is shown only if any host field is not `*`.

Also when I tested this first it printed nothing with zero kafka-native acls. Maybe somehow print always something even if no kafka-native acls are set.

tvainika, line 6264: For readability resrouce type should be first (left side) of resource name

aiven/client/cli.py
biggusdonzus: Going with this i modified this argument to be `--deny`, if not specified it is allow.

aiven/client/cli.py
biggusdonzus: :+1: for the Host, about the print with zero acls i'll think about it

biggusdonzus: Now i've aligned this with the UX thread

aiven/client/cli.py
tvainika: This is argparse based so is it possible to add mutually exclusive group as in https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group ?

tvainika (approved):

PR #403, opened 2025-03-12
Add support for Aiven for OpenSearch Custom Repo API

axeoman:

Add support for Aiven OpenSearch Custom Repo API, which includes listing custom repositories in OpenSearch and creating and restoring snapshots with it.

bonus: refactoring of cli/client modules by splitting it to base/common classess and making them extendable with service specific commands/endpoints


axeoman:
We are going to change endpoints in the Aiven API in order to mimic OpenSearch API so itself e.g. adding a _snapshot in path, so this should be changed in client part as well. Marking this PR as a draft until we get final list of endpoints.

axeoman:
Changes just got merged, so this is also ready to go https://github.com/aiven/aiven-core/pull/57763

PR #411, opened 2025-05-12
Use default project's account-id to check service versions availability

bugant:

It is possible to configure EOL extensions and also to enable specific versions for selected customers on Aiven backend. 
 
When creating or updating a service, check for service versions availability including also any account-level configurations.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -4025,7 +4025,7 @@ ssl.truststore.type=JKS
         requested_version = self._extract_user_config_version(service_type, user_config)
 
         if requested_version:
-            self._do_version_eol_check(service_type, requested_version)
+            self._do_version_eol_check(service_type, requested_version, project=project)
 
         service_integrations = []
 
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -4101,8 +4101,19 @@ def _get_service_type_user_config_schema(self, project: str, service_type: str)
 
         return service_def["user_config_schema"]
 
+    def _get_service_versions(self) -> Sequence[Mapping[str, Any]]:
+        default_project = self.config.get("default_project")
```
jeqo, line 4105: Tests seem to be failing as default project may not exist. We may need to check that the attribute exist first.
Also, should we consider adding an argument to pass a project instead of switching to check versions?

bugant: `'AivenCLI' object has no attribute 'config'` :exploding_head:  I thought that was not possible.  I'll fix that.

Prime541 (approved):

PR #413, opened 2025-05-16
feat: add inkless.enable config to topic creation

jeqo:
Inkless topics can be enabled only on creation.
Adding the config flags to support this.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -155,6 +155,44 @@ def test_service_user_create() -> None:
                 "tags": [],
             },
         ),
+        (
+            (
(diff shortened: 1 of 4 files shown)


tests/test_cli.py
```diff
@@ -155,6 +155,41 @@ def test_service_user_create() -> None:
                 "tags": [],
             },
         ),
+        (
+            ("service topic-create --project project1 --partitions 1 --replication 1 " + "--inkless-enable service1 topic1"),
```
juha-aiven, line 159: The concatenation can be removed.

juha-aiven (approved):
LGTM

PR #414, opened 2025-06-12
feat: add inkless.enable config to topic update

davide-armand:

Updating inkless topics fails because the client does not allow providing `inkless.enable`, so it is defaulted to False and inkless returns:
400 - Invalid config for topic <topic name>: It is invalid to disable Inkless


juha-aiven (commented):
This looks correct but could you please add also the CLI change and test similarly as in #413?

davide-armand:
> This looks correct but could you please add also the CLI change and test similarly as in #413?

Added now.

On a second thought, it looks a little unnecessary to have to provide `inkless.enable` on any config update, given that it is not possible to update it anyway (cannot turn an inkless topic into a non-inkless one and vice versa).
I wonder if, instead, this should be solved on inkless side: inkless could just keep the pre-existing value of `inkless.enable` when not provided in the config update (and keep returning error when trying to update it, as it already does).

juha-aiven:
> On a second thought, it looks a little unnecessary to have to provide `inkless.enable` on any config update, given that it is not possible to update it anyway (cannot turn an inkless topic into a non-inkless one and vice versa). I wonder if, instead, this should be solved on inkless side: inkless could just keep the pre-existing value of `inkless.enable` when not provided in the config update (and keep returning error when trying to update it, as it already does).

Right. I see what mean and it's a valid thing to consider.

However, I think it's correct in the case of inkless topic to require `inkless.enable=true` in any topic config update. This is because the semantics of the API call is replace the current topic config with the given one (if they differ). Not requiring the `inkless.enable` would change the semantics for one config key. I know it's ideal, but it's aligned with the API semantics.

juha-aiven (approved):
LGTM

PR #419, opened 2025-07-10
Add commands to manage permissions

bugant:
Add commands to configure and list permissions associated to a given resource.

Note the `permissions set` command operate in `set` mode, so it will always override the total of the configured permissions.  Also, it is not possible to work with permissions on project and organization level on the same command.

The new commands can be used to configure application-user permissions, for example, as it follows:

```
➜ avn permissions list --organization-id org5478b1b3c91 --resource-type organization                                                                                                                                                                                                                                           11:28:06 [83/1777]
No permissions found

➜  avn permissions set --organization-id org5478b1b3c91 --resource-type organization --principal-id u5478b350907 --principal-type user --permissions 'organization:projects:write' --permissions 'organization:billing:read'
The following permissions are going to be added:
+ organization:projects:write
+ organization:billing:read
Do you want to proceed (y/N)?y
Done

➜ avn permissions list --organization-id org5478b1b3c91 --resource-type organization 
PRINCIPAL_ID  PRINCIPAL_TYPE  PERMISSIONS                                            
============  ==============  ====================================================== 
u5478b350907  user            organization:billing:read, organization:projects:write 
u5478b87cd37  user            role:organization:admin

➜ avn permissions list --organization-id org5478b1b3c91 --resource-type organization --principal-id u5478b87cd37
PRINCIPAL_ID  PRINCIPAL_TYPE  PERMISSIONS            
============  ==============  =======================
u5478b87cd37  user            role:organization:admin

➜ avn permissions list --organization-id org5478b1b3c91 --resource-type organization --principal-type user
PRINCIPAL_ID  PRINCIPAL_TYPE  PERMISSIONS                                            
============  ==============  ====================================================== 
u5478b350907  user            organization:billing:read, organization:projects:write 
u5478b87cd37  user            role:organization:admin
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -192,6 +192,24 @@ View project management event log::
 
 .. _services:
 
+Working with Permissions
+-----------------------
+
+List organization permissions for a given resource type (one of "organization", "organization_unit", or "project")::
+
+  $ avn permissions list --organization <organization_id> --resource-type <resource_type>
+
+Set permissions for a given principal in the scope of resource type and resource ID::
+
+  $ avn permissions set --organization <organization_id> --resource-type <resource_type> --resource-id <resource_id> --principal-id <principal_id> --principal-type <principal_type> --permission <permission1> --permission <permission2>
+
+Note that setting permissions will override any existing permissions for the specified resource type and resource ID.
+
+For example the following commands will set the permissions for the project with ID "proj1" to assign role "developer" to the user with ID "user1", and let them to be an "operator" for project "proj2"::
+
+  $ avn permissions set --organization <organization_id> --resource-type project --resource-id proj1 --permission developer --principal-id user1 --principal-type user
+  $ avn permissions set --organization <organization_id> --resource-type project --resource-id proj2 --permission operator --principal-id user1 --principal-type user
+
 Explore Existing Services
 -------------------------
 


aiven/client/cli.py
```diff
@@ -5000,6 +5000,89 @@ def application_user__delete(self) -> None:
         self.client.delete_application_user(organization_id=self.args.organization_id, user_id=self.args.user_id)
         print(f"Application user {self.args.user_id} deleted successfully.")
 
+    @arg.json
+    @arg.force
+    @arg.organization_id
+    @arg("--resource-type", choices=["organization", "project"], help="Resource type", required=True)
+    @arg("--project-id", help="Project ID, required if resource_type is 'project'")
```
ngilles-aiven, line 5007: Does it not make sen to leave this a bit more generic (and support `organization_unit`):
```suggestion
    @arg("--resource-type", choices=["organization", 'organization_unit", "project"], help="Resource type", required=True)
    @arg("--resource-id", help="Project ID, required if resource_type is not 'organization'")
```

as the value is already collapsed below as:
```python
resource_id = self.args.organization_id if self.args.resource_type == "organization" else self.args.project_id
```

aiven/client/cli.py
```diff
@@ -5000,6 +5000,89 @@ def application_user__delete(self) -> None:
...
+    @arg(
+        "--permissions",
+        action="append",
+        default=[],
+        help="Permissions to assign to the application user. Can be specified multiple times to add multiple permissions",
+    )
+    @arg("user_id", help="Application User ID")
+    def application_user__permissions__set(self) -> None:
```
ngilles-aiven, line 5015: The API invoked is not limited to application users (which from this perspective are just like normal users), it making the command just `permissions__set` and allowing `--principal-type` and `--principal-id` to allow setting these for users and groups.

ngilles-aiven (dismissed):
LGTM as is, but as the API is generic, I think it may be worth making the commands a bit more generic (though we can of course add a some smarts as needed, eg. doing project look ups by name, as I don't know how "easy" it is to grab the project id for customer)

Thoughts?

aiven/client/cli.py
```diff
@@ -5104,6 +5104,110 @@ def application_user__token__revoke(self) -> None:
...
+        if not self.args.permissions and not self.args.force:
+            self.print_boxed(
+                [
+                    f"You are going to remove all permissions for principal {self.args.principal_id} (principal_type={self.args.principal_type}).",
+                ]
+            )
+            if not self.confirm("Do you want to proceed (y/N)?"):
+                raise argx.UserError("Aborted")
+
+        resource_id = self.args.organization_id if self.args.resource_type in {"organization", "organization_unit"} else self.args.project_id
```
giacomo-alzetta-aiven, line 5134: Here it mentions `organization_unit` but that's not a valid option in the choices

giacomo-alzetta-aiven (commented):
So my understanding is that each call to update can only "set" the permission for a single project/organization but using multiple calls, one per organization/project you can set different permissions for multiple organizations/projects. 

I think we should also update the README and include at least one example with different permissions with multiple orgs/projects just to show how to properly use it.

bugant:
> So my understanding is that each call to update can only "set" the permission for a single project/organization but using multiple calls, one per organization/project you can set different permissions for multiple organizations/projects.
> 
> I think we should also update the README and include at least one example with different permissions with multiple orgs/projects just to show how to properly use it.

I added a minimal doc for these command to the README. I did not add much more examples, I think this should be clear, as it works very similar to the APIs.

ali-gholami-aiven (approved):

PR #420, opened 2025-07-11
Add command to show info for an application user's token

bugant:
It might be useful to retrieve information on a given application-user's access token, given it's token-prefix. This can be used, for example, to check when the token was last used, or its expire date.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -5082,6 +5082,43 @@ server_encryption_options:
         self.print_response(tokens, json=self.args.json, table_layout=layout)
 
     @arg.json
+    @arg.organization_id
+    @arg("--verbose", action="store_true", default=False, help="Show more details")
+    @arg("user_id", help="Application User ID")
+    @arg("token_prefix", help="Token prefix to get info for")
+    def application_user__token__info(self) -> None:
+        """Show info on the specified token for an application user."""
+        tokens = self.client.list_application_user_tokens(
+            organization_id=self.args.organization_id,
+            user_id=self.args.user_id,
+        )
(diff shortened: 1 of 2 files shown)


aiven/client/cli.py
```diff
@@ -5081,6 +5081,43 @@ def application_user__token__list(self) -> None:
             )
         self.print_response(tokens, json=self.args.json, table_layout=layout)
 
+    @arg.json
+    @arg.organization_id
+    @arg("--verbose", action="store_true", default=False, help="Show more details")
+    @arg("user_id", help="Application User ID")
+    @arg("token_prefix", help="Token prefix to get info for")
+    def application_user__token__info(self) -> None:
+        """List all tokens for an application user."""
```
jadkik, line 5090: The docstring needs to be updated

aiven/client/cli.py
```diff
@@ -5081,6 +5081,43 @@ def application_user__token__list(self) -> None:
...
+        tokens = self.client.list_application_user_tokens(
+            organization_id=self.args.organization_id,
+            user_id=self.args.user_id,
+        )
+        token = next((token for token in tokens if token["token_prefix"] == self.args.token_prefix), None)
```
jadkik, line 5095: Is it possible to have two tokens with the same prefix? Sounds unlikely but if it's a possible edge case, it would be good to print a warning of fail loudly IMO.

giacomo-alzetta-aiven: My understanding is that we use the token_prefix as identifier in the API (e.g. to revoke a token)... I assume that our server-side code ensures this must be unique... if not we should fix it there.

giacomo-alzetta-aiven (approved):
LGTM

PR #433, opened 2025-11-26
client: support cmk operations

hnousiainen:
Implement initial support for CMK operations.

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -6995,6 +6995,128 @@ server_encryption_options:
         ]
         self.print_response(rates, json=self.args.json, table_layout=layout)
 
+    @arg.project
+    @arg.json
+    def project__cmks__accessors(self) -> None:
+        """List project customer managed key accessors"""
+        response = self.client.project_cmks_accessors(
+            project=self.get_project(),
+        )
+        if self.args.json:
+            self.print_response(response, json=self.args.json)
+        else:
+            # Using custom formatting here, as accessor fields could change
+            print("PROVIDER  DETAILS")
+            print("========  =======")
+            for provider in response["accessors"]:
+                details = f"\n{'':8}  ".join(f"{k}  {v}" for k, v in response["accessors"][provider].items())
+                print(f"{provider:8}  {details}")
+
+    @arg.project
+    @arg.json
+    def project__cmks__list(self) -> None:
+        """List project customer managed keys (CMKs)"""
+        response = self.client.project_cmks_list(
+            project=self.get_project(),
+        )
+        if self.args.json:
+            self.print_response(response, json=self.args.json)
+        else:
+            layout = ["id", "status", "provider", "resource", "default_cmk", "created_at"]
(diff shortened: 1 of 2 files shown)


glenn-mcallister-aiven (dismissed):
I have to block this, for a couple of reason:

1. Most of these commands are using optional style arguments (i.e., `--cmk-id`) for required arguments. That's a major UX no-no.
2. WAY MORE significantly, the current implementation is intended for xxxxxxx only, to the point where
    1. we are hiding access to the the API behind an ACL, 
    2. we aren't documenting the API in our public API docs,
    3. xxxxxx does changes only through TF, meaning they need API access only.

Once these changes get published, customers are going to see that there are "customer managed keys" commands, which is 100% what we don't want currently.

aiven/client/cli.py
```diff
@@ -6995,6 +6995,128 @@ def inkless__offering__rates(self) -> None:
...
+                for k, v in response["cmk"].items()
+            ]
+            self.print_response(fields, json=self.args.json, table_layout=layout)
+
+    @arg.project
+    @arg("--cmk-id", help="CMK Identifier", required=True)
+    @arg("--default-cmk", help="Mark CMK as default for all new services", choices=["true", "false"])
+    @arg.json
+    def project__cmks__update(self) -> None:
+        """Define a project customer managed key (CMK)"""
```
VadimPushtaev, line 7081: Copy-pasted message, probably should be "Change", not "Define".

aiven/client/cli.py
```diff
@@ -6995,6 +6995,128 @@ def inkless__offering__rates(self) -> None:
...
+            self.print_response(response, json=self.args.json)
+        else:
+            layout = ["id", "status", "provider", "resource", "default_cmk", "created_at"]
+            self.print_response(response["cmks"], json=self.args.json, table_layout=layout)
+
+    @arg.project
+    @arg("--cmk-id", help="CMK Identifier", required=True)
+    @arg.json
+    def project__cmks__get(self) -> None:
+        """Get details on a customer managed keys (CMKs)"""
```
VadimPushtaev, line 7032: *key

aiven/client/cli.py
```diff
@@ -6995,6 +6995,128 @@ def inkless__offering__rates(self) -> None:
...
+                }
+                for k, v in response["cmk"].items()
+            ]
+            self.print_response(fields, json=self.args.json, table_layout=layout)
+
+    @arg.project
+    @arg("--cmk-id", help="CMK Identifier", required=True)
+    @arg.json
+    def project__cmks__delete(self) -> None:
+        """Delete a customer managed keys (CMKs)"""
```
VadimPushtaev, line 7109: * keys

VadimPushtaev (changes requested):
Some typos, but generally fine. `--cmk-id` being required pattern is a pretty common pattern for aiven-client anyway.

hnousiainen:
Closing in favor of https://github.com/aiven/aiven-client/pull/451

PR #450, opened 2026-04-01
client: fix typo in disk space gib argument metavar

jadkik:

The help test reads `--disk-space-gib DISK_SPACE_MB` in the first
line and lower in the description:

```
  --disk-space-gib DISK_SPACE_MB
                        Disk space for data storage (GiB)
```

The metavar `DISK_SPACE_MB` is inferred from the variable name since we
implicitly convert to MiB by multiplying by 1024 in the code. By
specifying it explicitly we can avoid confusing the user as they'll not
be exposed to the MiB value, neither in the usage of the CLI or in the
error message.

It's only relevant if they are making the API call directly.

### DIFF
--- a/aiven/client/cliarg.py
+++ b/aiven/client/cliarg.py
@@ -116,7 +116,11 @@ arg.config_file = arg(
 )
 arg.country_code = arg("--country-code", help="Billing country code")
 arg.disk_space_mb = arg(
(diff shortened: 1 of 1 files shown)


giacomo-alzetta-aiven (approved):
LGTM

PR #454, opened 2026-05-08
replace black, isort, and flake8 with ruff

dogukancagatay:
Ruff was already used as a linter, its formatter and isort rules can
fully replace the three separate tools. This shrinks the dev toolchain
to a single fast tool with one config block in pyproject.toml.

Files are reformatted. One file touched by `ruff format` and nine
`from x import a, b, c` lists re-sorted to ruff's member order.


szypulka (approved):
LGTM

PR #464, opened 2026-06-19
Treat empty --user-config-json ("{}") as an explicit empty config

dogukancagatay:
A missing flag and an explicit `"{}"` both collapsed to `{}`, and the
truthy call-site checks treated the empty dict as "not provided",
silently ignoring an intentionally-empty config. Gate parsing and all
call sites on `is not None` so an explicit empty config is honored.


ezotrank (approved):
lgtm

PR #466, opened 2026-06-22
client: bound HTTP keepalive connection reuse to mitigate timeout race

mbertheau:
## Summary
- Recycle pooled HTTP keepalive connections in `AivenClientAdapter` before they outlive the API server's connection window, so the client never writes a request onto a connection the server has already closed (which surfaces as a fatal `RemoteDisconnected` and cannot be safely retried for non-idempotent methods).
- Two pool-wide bounds, checked before each request and enforced via the public `urllib3` `PoolManager.clear()` (no private APIs):
  - idle timeout (default 50s): recycle a connection idle longer than this.
  - max age (default 59m): recycle a connection older than this regardless of activity.
- Configurable via `get_requests_session()`, `AivenClient()`, and new `--idle-timeout` / `--max-age` CLI flags.
- Add tests for the new defaults and parameter passthrough.

## Notes
- Library API (`get_requests_session`, `AivenClient`) stays backwards-compatible; the only break is the internal `ClientFactory` extension point, which gains two parameters.
- Per-connection timers (like nginx-style idle/max-age) would require subclassing private `urllib3` pool internals; the pool-wide approach trades that fragility for coarser recycling, which is practically equivalent for this single-host, mostly-sequential CLI.

## Test plan
- [x] `make lint` (ruff check + format + mypy) passes.
- [x] `make test` passes (incl. new `tests/test_session.py` cases).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -7,14 +7,14 @@ from __future__ import annotations
 from . import argx, client, units
 from aiven.client import AivenClient, envdefault
 from aiven.client.cliarg import arg
-from aiven.client.client import Tag
+from aiven.client.client import DEFAULT_IDLE_TIMEOUT, DEFAULT_MAX_AGE, Tag
 from aiven.client.common import UNDEFINED
 from aiven.client.connection_info.common import Store
 from aiven.client.connection_info.kafka import KafkaCertificateConnectionInfo, KafkaSASLConnectionInfo
 from aiven.client.connection_info.pg import PGConnectionInfo
 from aiven.client.connection_info.valkey import ValkeyConnectionInfo
 from aiven.client.speller import suggest
-from argparse import ArgumentParser
+from argparse import ArgumentParser, ArgumentTypeError
 from ast import literal_eval
 from collections import Counter
 from collections.abc import Callable, Mapping, Sequence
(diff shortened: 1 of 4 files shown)


aiven/client/cli.py
```diff
@@ -174,6 +182,18 @@ def add_args(self, parser: ArgumentParser) -> None:
...
+            type=float,
+            default=DEFAULT_IDLE_TIMEOUT,
+            help="Recycle a pooled HTTP connection idle for more than N seconds (default: %(default)s)",
+        )
+        parser.add_argument(
+            "--max-age",
+            type=float,
+            default=DEFAULT_MAX_AGE,
+            help="Recycle a pooled HTTP connection older than N seconds (default: %(default)s)",
+        )
```
jadkik, line 205: nit: The way the arguments are defined here, they can never be `None` when passed down the chain of functions. So we can't really disable this like we could by calling the internal classes directly.

I believe that's not a big deal because the user can also just raise the values to absurd amounts and then it's almost equivalent.

jadkik (dismissed):
LGTM - only a nit comment but I think we can merge anyway if you agree

aiven/client/cli.py
mbertheau: Yeah, I'll have -1 mean disabled.

jadkik (approved):

PR #478, opened 2026-09-08
Migrate org VPC user-peer-network-cidrs to new API [AI-601]

rsalevsky:
## Summary

- Migrate `organization vpc peering-connection user-peer-network-cidrs add/delete` to the VPC-level replacement API.
- Update `organization_vpc_user_peer_network_cidrs_update()` to call `PUT /v1/organization/{org_id}/vpc/{vpc_id}/user-peer-network-cidrs` instead of the sunset per-peering-connection path.
- Remove `--peering-connection-id` from the CLI; the old path parameter was ignored by the server.

The per-peering-connection endpoint was sunset on 2026-04-30. Request body semantics are unchanged (`add` / `delete`).

## Test plan

- [x] `python3 -m pytest tests/ -q` (242 passed)
- [x] `python3 -m pytest tests/test_cli.py -k organization_vpc_peering_connection_user_peer -q`

Jira: https://aiven.atlassian.net/browse/AI-601

### DIFF
--- a/aiven/client/cli.py
+++ b/aiven/client/cli.py
@@ -5754,29 +5754,35 @@ ssl.truststore.type=JKS
 
     @arg("--organization-id", required=True, help="Identifier of the organization of the custom cloud environment")
     @arg("--organization-vpc-id", required=True)
-    @arg("--peering-connection-id", required=True)
+    @arg("--peer-cloud-account", required=True, help=_peer_cloud_account_help)
+    @arg("--peer-resource-group", help=_peer_resource_group_help)
+    @arg("--peer-vpc", required=True, help=_peer_vpc_help)
     @arg("cidrs", nargs="+", metavar="CIDR")
     @arg.json
     def organization__vpc__peering_connection__user_peer_network_cidrs__add(self) -> None:
-        """Add user peer network CIDRs for an organization VPC peering connection."""
+        """Add user peer network CIDRs for an organization VPC."""
+        add_base = {
+            "peer_cloud_account": self.args.peer_cloud_account,
+            "peer_vpc": self.args.peer_vpc,
+        }
+        if self.args.peer_resource_group is not None:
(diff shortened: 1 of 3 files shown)


aiven/client/cli.py
```diff
@@ -6123,29 +6123,25 @@ def organization__vpc__clouds__list(self) -> None:
...
     @arg("cidrs", nargs="+", metavar="CIDR")
     @arg.json
     def organization__vpc__peering_connection__user_peer_network_cidrs__add(self) -> None:
-        """Add user peer network CIDRs for an organization VPC peering connection."""
+        """Add user peer network CIDRs for an organization VPC."""
         self.client.organization_vpc_user_peer_network_cidrs_update(
             organization_id=self.args.organization_id,
             organization_vpc_id=self.args.organization_vpc_id,
-            peering_connection_id=self.args.peering_connection_id,
             add=[{"cidr": cidr} for cidr in self.args.cidrs],
```
dogukancagatay: According to the public API docs, update endpoint requires `peer_cloud_account` and `peer_vpc` along with `cidr`.

https://api.aiven.io/doc/#tag/Organization_Vpc/operation/OrganizationVpcPeeringConnectionPeerNetworkCidrsUpdateV2

aiven/client/cli.py
```diff
@@ -5754,29 +5754,35 @@ def organization__vpc__clouds__list(self) -> None:
 
     @arg("--organization-id", required=True, help="Identifier of the organization of the custom cloud environment")
     @arg("--organization-vpc-id", required=True)
-    @arg("--peering-connection-id", required=True)
```
d812k, line 5757: this is a hard change, if there's a script that still uses this parameter, it will fail with an error. May be adding a deprecation notice instead and then removing in the next release?

d812k (approved):
lgtm

